PR #3, opened 2022-02-01
WIP: Add GetAccountBlockByMomentumAcknowledged to rpc/api/ledger.go

georgezgeorgez:
Uses binary search to get the latest AccountBlock for an address which was acknowledged by a certain Momentum height.
Not sure how useful this is on its own but this can help get the state of the Account at that Momentum height.
It's possible for this all to be done by the client, but having it implemented on the node shouldn't add operational overhead to the node.

TODO: cleanup, exposing through the rpc if not automatic, tests

### DIFF
--- a/rpc/api/ledger.go
+++ b/rpc/api/ledger.go
@@ -291,6 +291,65 @@ func (l *LedgerApi) GetUnreceivedBlocksByAddress(address types.Address, pageInde
 	}, nil
 }
 
+// Glue
+
+func (l *LedgerApi) GetAccountBlockByMomentumAcknowledged(address types.Address, momentumHeight uint64) (*AccountBlock, error) {
+	l.log.Info("GetAccountBlockByMomentumAcknowledged")
+
+	accountStore := l.chain.GetFrontierAccountStore(address)
+	frontierAccountBlock, err := accountStore.Frontier()
+	if err != nil {
+		l.log.Error("GetFrontierAccountBlock failed, error is "+err.Error(), "method", "GetAccountBlockByMomentumAcknowledged")
+		return nil, err
+	}
+	if frontierAccountBlock == nil {
+		return nil, nil
+	}
+
+	// check because already retrieved
+	fMomentumAcknowledged := frontierAccountBlock.MomentumAcknowledged.Height
+	if fMomentumAcknowledged <= momentumHeight {
+		return ledgerAccountBlockToRpc(l.chain, frontierAccountBlock)
+	}
+
+	// TODO: Remove commented, in for clarity right now
+	//	genesisAccountBlock, err := accountStore.ByHeight(1)
+	//	if err != nil {
+	//		l.log.Error("GetGenesisAccountBlock failed, error is "+err.Error(), "method", "GetAccountBlockByMomentumAcknowledged")
+	//		return nil, err
+	//	}
+	//
+	//	gMomentumAcknowledged := genesisAccountBlock.MomentumAcknowledged.Height
+	//	if gMomentumAcknowledged > momentumHeight {
+	//		return nil, nil
+	//	}
+
+	// binary search for latest block where momentum acknowleged <= desired momentum height
+	var block *nom.AccountBlock = nil
+	low := uint64(1)
+	high := frontierAccountBlock.Height
+
+	for low <= high {
+		mid := low + (high-low)/2
+		midBlock, err := accountStore.ByHeight(mid)
+		if err != nil {
+			l.log.Error("GetAccountBlockByHeight failed, error is "+err.Error(), "method", "GetAccountBlockByMomentumAcknowledged")
+			return nil, err
+		}
+		if midBlock.MomentumAcknowledged.Height <= momentumHeight {
+			block = midBlock
+			low = mid + 1
+		} else {
+			high = mid - 1
+		}
+	}
+
+	if block == nil {
+		return nil, nil
+	}
+	return ledgerAccountBlockToRpc(l.chain, block)
+}
+
 // Momentum
 func (l *LedgerApi) GetFrontierMomentum() (*Momentum, error) {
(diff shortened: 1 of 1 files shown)


dumeriz:
A comment on the implementation only: Have you checked the binary search algorithm from go's stdlib?
Maybe you have and it's not applicable for the problem at hand. If so, can you explain why?
Otherwise, when it's available from acknowledged libraries, I wouldn't implement such low level stuff again. 
It just adds unnecessary complexity and possible error sources.

https://pkg.go.dev/sort

georgezgeorgez:
lol i knew it existed somewhere and used it long ago but couldn't find it, thanks
i was looking for a 'search' or 'algorithm' package, but makes sense that its with sort
shows the extent of my current fluency with go...
i guess the only possible argument against using it would be that we not only want the index but the block itself
actually i started working on this by returning an integer index, but figured to change it the return the block itself
1. to fit in with the existing rpc return types
2. the account chain height is mostly useless on its own, the first thing the caller would do is probably fetch the block by height
3. we have to fetch the block anyways to do the search so with 2), it makes sense just to return the block
the same consideration could apply when it comes to using sort.Search
a simple predicate function to fetch the account block, test it, and then return the bool would mean that before returning final account block, another call to the account store would be needed
but much cleaner code, and that extra call is probably whatever
maybe i can use a closure to avoid the final call

georgezgeorgez:
[sort.] Search uses binary search to find and return the smallest index i in [0, n) at which f(i) is true

I don't think it's like cpp with nice lower_bound and upper_bound search
I'm looking for the greatest index

I guess I can -1
The std library example mentions reversing a list to invert the condition, but not working with a list

Even before implementation though, this is a rational thing to search for right?

georgezgeorgez:
Also 0 index vs 1 index

georgezgeorgez:
```
$ make znnd
go build -o /home/george/repos/zenon-misc/go-zenon/build/znnd /home/george/repos/zenon-misc/go-zenon/main.go
# github.com/zenon-network/go-zenon/rpc/api
rpc/api/ledger.go:317:40: cannot use n + 1 (type int) as type uint64 in argument to accountStore.ByHeight
rpc/api/ledger.go:320:4: too many arguments to return
	have (nil, error)
	want (bool)
rpc/api/ledger.go:326:23: cannot use high (type uint64) as type int in argument to sort.Search
rpc/api/ledger.go:336:37: cannot use found (type int) as type uint64 in argument to accountStore.ByHeight
make: *** [Makefile:9: znnd] Error 2

```

georgezgeorgez:
rpc/api/ledger.go:320:4: too many arguments to return
	have (nil, error)
	want (bool)
	
This is just me being dumb with copy and paste and can be fixed. Can pass the error through closure
Not sure if there's a way around the uint64 type

dumeriz:
I guess the version matching your implementation would look something like this:
```
max := frontierAccountBlock.Height
var block_std *nom.AccountBlock = nil
idx := sort.Search(max+1, func(i int) bool {
    block, err := accountStore.ByHeight(i)
    if err != nil {
        l.log.Error("GetAccountBlockByHeight failed, error is "+err.Error(), "method", "GetAccountBlockByMomentumAcknowledged")
        return false
    }
    return block.MomentumAcknowledged.Height <= momentumHeight
})

if idx <= max && block_std != nil {
    return ledgerAccountBlockToRpc(l.chain, block_std)
}

return nil, nil
```
But yeah, that fails because of the fixed types and the missing generic interface. That's astonishing.

Anyway, here's the [implementation](https://go.dev/src/sort/search.go). Maybe it would be beneficial to copy this for uint64 and then maybe other basic types as needed, and put them into a dedicated unit, maybe in the `common` package. It can also be more easily tested then.

Grep'ing for search in the code shows that some form of binary search is used in at least 2 other locations.

georgezgeorgez:
afaik
zenon account block are [1, height]
sort.search is [0, n)
so your +1 just makes it [0,n] or [0,n+1)

what i did was use [0, n)
but inside the predicate function use i + 1
effectively giving me [1, n+1) or [1,n]

also sort search returns the first index where the predicate is true
we want the last index
so i have to invert the predicate
the first account block where its acknowledged height > momentum height we specify
and then take the index before that

also probably don't want the predicate function to return false if it has an error
a bit tricky, but probably want to use an external err variable, and have the function be a closure which writes to that if there's an error
then once the search is done, check the err and return nil, err if err != nil

georgezgeorgez:
if the found index is 0, it means that the first account block satisifies our inverted predicate
meaning that the chain had no blocks at our specified momentum height
in this case i return nil, nil

if the found index is -1
it means none of the blocks satisfy our inverted predicate
all blocks are acknowledged before our specified momentum height
in this case, we return the frontierBlock

and actually in my code specifically, i check the frontier before doing the binary search because we already retrieve it
so we'll never get a -1

if we get an index > 0, then we take first subtract 1 because of the inverted condition, but then add 1 because of the 1 index
so we return accountStore.byHeight(index)

georgezgeorgez:
so it's either implement binary search directly like i have which takes into consideration the 1 index and that we want the greatest index which satisifes a condition
or use a uint64 variant of sort.Search and reason about the workarounds

dumeriz:
The indices could be adjusted accordingly. It was more or less a quick test to understand the problem when you said you return both index and block. 
Apparently go lambdas close over the scope, so we can access the block var from inside the predicate. Same would hold for the err variable, and if you wanted, you could always return false once err has been set. 
And true, the predicate would have to be reversed. 

I think it's preferable to adjust a bit, if it in turn allows to extract the lower level stuff. 

Maybe that's not the way to go though (pun intended), and it would be much nicer if generics were stable already for sure.

georgezgeorgez:
hopefully 1.18 makes things a bit easier

i thought github allowed me to change the source branch of PRs
apparently not
going to close this and reopen this later with some updates and proper branching

PR #8, opened 2022-03-01
Updated bech32 dependency URL

oxmah:
github.com/btcsuite/btcutil/bech32 moved to github.com/btcsuite/btcd/tree/master/btcutil/bech32


georgezgeorgez:
I'm going to close this as stale
I suggested this as an easy PR, but updating dependencies requires updating the go.mod & go.sum files
And we would probably want to run some tests.

PR #10, opened 2022-05-30
WIP: Add generate-devnet command

georgezgeorgez:


zenon-network:
Great initiative! What's the progress on this?

georgezgeorgez:
It's usable afaik
Needs some cleanup for coding standards etc

znnd generate-devnet --genesis-block=<address>,<tZNN amount>,<tQSR amount>

The --genesis-block flag can be used multiple times and it handles the decimals conversion

Existing --data flag works of course meaning you can create and delete local devnets easily
Other existing flags should also work just fine

I want to add some more flags to allow people to choose ProtocolVersion and ChainIdentifier inside the generated genesis file 

Maybe a --genesis-fusion flag 

And maybe something to generate a config file, without the producer, that takes in an --ip flag, to create a config with seeder than can be shared

georgezgeorgez:
I'm not really sure if this kind of command makes sense to bundle with the node code. It could be done as a separate utility. But this seemed to be the fastest thing to start with so that developers can create local developer environments. Making it a command seemed to be the least intrusive.

georgezgeorgez:
Need to fix a bug causing it to crash on height 181 
Guessing every 30 minutes
There is a peer discovery 
And running into an indexing error
Because my peer list is empty
I think I can add localhost to the seeders and it'll be okay

georgezgeorgez:
The aforementioned bug was fixed

georgezgeorgez:
decided a while back that I was going to move this functionality into nomctl

PR #17, opened 2023-04-03
NoM Multi-chain Technology and Orbital Program support

nomkamoto:


nomkamoto (commented):
Added the required validation for the `setTokenPair` method.

PR #20, opened 2023-05-15
big-int-to-string

MoonBaZZe:
All RPC responses will return big.Int as string to increase compatibility with high decimal tokens. Required difficulty was converted to uint64 because it cannot overflow it, even though it was big.Int


sumamu:
@zenon-network this functionality is necessary and not breaking.

georgezgeorgez:
there are some better ways to do this
hold on

georgezgeorgez:
So at the very least we should refactor it like this
https://calvinfeng.gitbook.io/gonotebook/idioms/custom-json-marshaling#better-approach
to reduce unnecessary lines of code

What I am evaluating as well is how invasive it would be to refactor with a custom type instead of big.Int

zenon-network:
@georgezgeorgez please submit a PR when you have a better implementation

georgezgeorgez:
sounds good!

0x3639:
I've run the latest release with the znn controller and get the following error when I deploy:

`type '_OneByteString' is not a subtype of type 'int'`

MoonBaZZe:
> So at the very least we should refactor it like this https://calvinfeng.gitbook.io/gonotebook/idioms/custom-json-marshaling#better-approach to reduce unnecessary lines of code
> 
> What I am evaluating as well is how invasive it would be to refactor with a custom type instead of big.Int

I also took this option into consideration when I chose to just implement Json marshal and unmarshal. It seemed to me that there would be much more code to implement with a new type.

MoonBaZZe:
Also the znn-cli is not working anymore and any other application that is not compiled with an sdk compatbile with big-int-string

MoonBaZZe:
Regarding the tests, on my local machine I ran them as bulk and every test passed. I guess on the CICD pipeline they just run every test as a single entity.

georgezgeorgez:
Custom json handling makes sense to me infrequently for special situations
Short term we gotta do what we gotta do to get the code working
And I would agree we are very much living in the short term right now

Long term it's an overhead tax to developers on all rpc returns
With the struct embedding method i linked, the tax is less

> It seemed to me that there would be much more code to implement with a new type.

```
go-zenon$ grep -rni big.Int | wc -l
282
```

In terms of LOC, I don't think it would be that much more
Would mostly need some wrapper functions so that NewType.Add(NewType) and other related functions would work
But otherwise, it should just be a drop in replacement

In terms of effort,
big.Int for amounts is so prevalent in the code
a refactor would be pretty invasive
so if anything went wrong, it might be a pain to debug

PR #22, opened 2023-07-14
Fix memory leaks

MoonBaZZe:
The leaks were created by spawning go routines on each setupConn or handle that would never returned. The Ethereum code base did not include these go routines in the releases so I removed them after thoroughly looking for other implications. I have tested it with a full resync and RPC.


Sol-Sanctum:
Several devs have[ reported positive results](https://forum.hypercore.one/t/pprof-results-from-go-zenon/127/11) for these changes.

Personally, though my node's memory utilization increases under load, it releases that memory slowly over time, instead of increasing slowly forever. 
No perceivable adverse impact.

PR #26, opened 2023-08-11
Fix fork

sumamu:
Waiting for more confirmations of successful resync from scratch.

### DIFF
--- a/.github/workflows/znn_builder.yml
+++ b/.github/workflows/znn_builder.yml
@@ -2,6 +2,8 @@ name: Build and release znnd and libznn
 
 on:
   push:
+    branches:
+      - master
   # Allows you to run this workflow manually from the Actions tab
   workflow_dispatch:
 
--- a/metadata/version.go
+++ b/metadata/version.go
@@ -3,5 +3,5 @@
 package metadata
 
 const (
-	Version = "v0.0.6"
+	Version = "v0.0.7"
 )
--- a/metadata/version_libznn.go
+++ b/metadata/version_libznn.go
@@ -3,5 +3,5 @@
 package metadata
 
 const (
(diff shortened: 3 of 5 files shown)


0x3639 (approved):
I've tested this on 5 different nodes with sync from bootstrap after the fork.  All were successful.  Other have tested from scratch and different bootstrap from before the fork.  All were successful.

georgezgeorgez:
I have yet to test this myself with a resync. 

I know I suggested stable sort as an option, but any particular reason why it was chosen over falling back to hashes in the case of time stamp collision?

Was it the fact that the sort stable "just worked" with the live chain?

Or did the the hash-based comparison fail to produce the live chain?

Just curious.

georgezgeorgez:
Also can we squash these commits?

georgezgeorgez:
After we fix this issue, we should proactively search the codebase for other sorts and see if it's a possible issue elsewhere.

sumamu:
> Was it the fact that the sort stable "just worked" with the live chain?
Fortunately, it did and this solution was more elegant than comparing hashes.

> Or did the the hash-based comparison fail to produce the live chain?
I didn't test this in the end. The other solution seemed better and it produced the expected results.

> The solution we choose is restricted by the need to produce the results that are already on chain.
There were 2 forks:
- one without any downtime
- the other with ~2  hours of downtime after the fork momentum.

We could've chosen any, but it seemed natural to choose the one without downtime because in the end it was the longer chain.

sumamu:
You're right to make this clear for future reference. 
I do remember there are other places in the go-zenon code where leveldb's lexicographical ordering is relied upon.

MoonBaZZe (approved):

georgezgeorgez (approved):
lgtm

PR #27, opened 2023-08-22
[rpc] Fix BalanceInfo UnmarshalJSON

georgezgeorgez:
Surprised that no one else ran into this issue yet...

### DIFF
--- a/rpc/api/ledger_types.go
+++ b/rpc/api/ledger_types.go
@@ -106,26 +106,14 @@ type BalanceInfo struct {
 }
 
 type BalanceInfoMarshal struct {
-	TokenMarshal `json:"token"`
-	Balance      string `json:"balance"`
+	TokenInfo *TokenMarshal `json:"token"`
+	Balance   string        `json:"balance"`
 }
 
 func (b *BalanceInfo) ToBalanceInfoMarshal() BalanceInfoMarshal {
 	aux := BalanceInfoMarshal{
-		TokenMarshal: TokenMarshal{
-			TokenName:          b.TokenInfo.TokenName,
-			TokenSymbol:        b.TokenInfo.TokenSymbol,
-			TokenDomain:        b.TokenInfo.TokenDomain,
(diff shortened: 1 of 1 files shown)


georgezgeorgez:
I'm running into errors related to unmarshalling BalanceInfo
Need to do some testing

georgezgeorgez:
I guess most clients are Dart (syrius) so aren't using these unmarshal methods

georgezgeorgez:
For AccountBlock, what is this check for?
`if block.TokenInfo != nil {`
in
```
func (block *AccountBlock) UnmarshalJSON(data []byte) error {
	aux := new(AccountBlockMarshal)
	if err := json.Unmarshal(data, aux); err != nil {
		return err
	}

	block.AccountBlock = *aux.FromNomMarshalJson()
	if block.TokenInfo != nil {
		block.TokenInfo = aux.TokenInfo.FromTokenMarshal()
	}
	block.ConfirmationDetail = aux.ConfirmationDetail
	if block.PairedAccountBlock != nil {
		block.PairedAccountBlock = aux.PairedAccountBlock.FromApiMarshalJson()
	}
	return nil
}
```

Is this checking the transaction is using the zero/empty token?
Wouldn't you check the aux then?
i.e `if aux.TokenInfo != nil {`

Making sure I don't need to do the same kind of check for BalanceInfo

georgezgeorgez:
My testing has confirmed this change works as intended

georgezgeorgez:
And I think I can confirm AccountBlock Unmarshal is not working properly also. I will create a separate PR for that.

Sol-Sanctum:
@sumamu @MoonBaZZe Can you please check this out?

MoonBaZZe (commented):
Seems good!

MoonBaZZe (approved):
Seems good!

sumamu (approved):

0x3639 (approved):

0x3639:
I've this question before, and someone indicated we should wait for the requestor to merge them.  So I was hoping George would do that.

PR #30, opened 2023-09-09
P2p/no seeders

georgezgeorgez:


georgezgeorgez:
Sigh, sorry I keep forgetting the default PR behavior is to merge to upstream

PR #31, opened 2023-09-09
[p2p] Allow discovery refresh to handle empty peers

georgezgeorgez:


georgezgeorgez:
This change is useful for single node testnets

georgezgeorgez:
Not sure if there are any stated assumptions about nonzero seeder length, but I will label this as a bug.
Either assert nonzero length, or check the length between doing arithmetic and logic on it.

MoonBaZZe (approved):
Nasty bug

0x3639:
@georgezgeorgez how did you discover this bug?  Is this related to the slow sync after fork, or discovery in setting up an single node testnet?

0x3639 (approved):

georgezgeorgez:
@0x3639 discovery in single node testnet
it would always crash after height 180
or after 30 minutes which is the frequency of the discovery refresh

sumamu (approved):

PR #34, opened 2023-09-22
[fmt] Omit comparison with boolean constant and fmt imports

georgezgeorgez:
https://staticcheck.dev/docs/checks/#S1002

### DIFF
--- a/chain/account_pool.go
+++ b/chain/account_pool.go
@@ -138,7 +138,7 @@ func (ap *accountPool) addAccountBlockTransaction(transaction *nom.AccountBlockT
 	if err := ap.canRollback(block); err != nil {
 		return err
 	}
-	if err := higherPriority(block, trueBlock); forceAdd == false && err != nil {
+	if err := higherPriority(block, trueBlock); !forceAdd && err != nil {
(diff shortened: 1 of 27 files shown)


0x3639 (approved):

georgezgeorgez:
to squash

MoonBaZZe:
Why did you have to do this change?

MoonBaZZe:
I suggest you do a full resync to test changes like this. I will approve after you tell me everything is ok

georgezgeorgez:
> Why did you have to do this change?

In terms of S1002 specifically, it's more idiomatic and less verbose.
Overall, static analysis is a useful tool if you are diligent about addressing it.
Always good to run checks on new code, but in order to get there need to maintain the old code to reduce noise.

> I suggest you do a full resync to test changes like this. I will approve after you tell me everything is ok

This is pretty much just a formatting change. The logic is the exact same.
It's possible I made an error, but that can be detected with review.

MoonBaZZe (approved):

sumamu (approved):

georgezgeorgez:
squashing

PR #35, opened 2023-09-22
Move getting git version into native build

georgezgeorgez:
https://stackoverflow.com/questions/71642366/how-do-you-read-debug-vcs-version-info-from-a-go-1-18-binary

Notably, that go commands are meant to run on package not file argument


MoonBaZZe (approved):

0x3639:
**Notes from my review:** The PR improves the way the git commit hash is retrieved and used in the build process. Instead of writing the commit hash to a file during the build process (using make version), the commit hash is now retrieved natively using Go's runtime/debug package. This approach is more dynamic and avoids the need to modify a Go file every time a new commit is made.

0x3639 (approved):

sumamu (approved):

PR #43, opened 2024-06-12
Improve account pool performance

vilkris4:
The account pool starts experiencing performance issues when it holds a large amount of account blocks from account chains that have long unconfirmed states. Specifically the account pool's (`account_pool.go`) functions `getUncommittedAccountBlocksByAddress` and `rebuild` become unacceptably slow if there are more than a few account chains with hundreds of pending account blocks on each chain.

The reason for the slowness is the multiple consecutive read operations done to the account chain's unconfirmed (frontier) database in the `getUncommittedAccountBlocksByAddress` and `rebuild` functions. The account chain's unconfirmed database's performance is not adequate for this use case.

To fix this issue, this commit introduces an `accountManager` which is responsible for maintaining a key-value map of an account's pending blocks. This way the pending blocks do not have to be fetched from the unconfirmed account database, significantly improving performance.

### Benchmarks
[This benchmark](https://github.com/vilkris4/go-zenon/blob/c59d17ac9c967f6c59dfe432122d1c68ebba4c59/chain/tests/account_pool_test.go#L123) was used to compare the performance of the current implementation to the new implementation. The benchmark adds a total of 2,500 account blocks from five different accounts into the account pool and then benchmarks the public function `GetAllUncommittedAccountBlocks`, which internally uses the problematic `getUncommittedAccountBlocksByAddress` function.

|Version|ns/op|ms/op|B/op|allocs/op|
|-----|-----|-----|-----|-----|
|Current (v0.0.7)|4476964800 ns/op|4476.96 ms/op|11485528 B/op|188399 allocs/op|
|New|4331860 ns/op|4.33 ms/op|5766849 B/op|95988 allocs/op|

As can be seen from the results, the current implementation becomes slow when there are a few hundred pending blocks for each account in the account pool.

### Testing

A full resync of the node has been tested with these changes.

### DIFF
--- a/chain/account_pool.go
+++ b/chain/account_pool.go
@@ -20,6 +20,7 @@ var (
 	ErrFailedToAddAccountBlockTransaction = errors.Errorf("failed to insert account-block-transaction")
 	ErrPlasmaRatioIsWorse                 = errors.Errorf("plasma ratio is smaller for current block")
 	ErrHashTieBreak                       = errors.Errorf("hash tie-break is worse for current block")
+	ErrBlockHeightNotFound                = errors.Errorf("block height does not exist in account manager")
 
 	// MaxAccountBlocksInMomentum takes into account batched account-blocks
 	MaxAccountBlocksInMomentum = 100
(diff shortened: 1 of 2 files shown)


vilkris4:
After further testing I have [removed](https://github.com/vilkris4/go-zenon/commit/844f69aba35757c0763f46cbaa32f581944bbf1e) the changes done to the account pool's rebuild functionality from this PR because of the following reason (copy of the commit message):

> In a situation where an account chain continuously has uncommitted blocks in the account pool, the "restabilization" of the `memdbManager` will eventually result in a massive amount of wrapped databases inside wrapped databases inside wrapped databases in the `memdbManager`'s `versions` map. This is because the `versions` map never gets fully cleared and rebuilt from scratch when the `Restabilize` method is used. These wrapped databases inside wrapped databases can cause significant issues in read performance when reading from the unconfirmed account database.
> 
> Because of this the changes to the rebuild functionality have been reverted, even though this issue is only an edge case. More consideration is needed on how to improve the account pool's rebuild performance, while minimizing unwanted side effects.

MoonBaZZe (approved):
I have reviewed the code after watching the youtube video and spending some time on it. It seems fine and does not mess with other logic as of my understanding of the code base.

0x3639:
For anyone interested in researching this PR here is a two part code review between George and Vilkris.  

Part 1/2
https://youtu.be/Pb87qKrtszs

Part 2/2
https://youtu.be/CZxMgTuKSuk

digitalSloth (approved):

0x3639:
**Findings**
- [P2] [chain/account_pool.go](/private/tmp/go-zenon-pr43.uukjsz/chain/account_pool.go:58) returns the exact block pointer stored in `accountManager.blocks`, and public getters append that pointer into caller-visible slices at [line 353](/private/tmp/go-zenon-pr43.uukjsz/chain/account_pool.go:353). Before this PR, `frontier.ByHeight` deserialized a fresh block from the DB, so callers could not mutate pool internals. Now a caller that changes a returned uncommitted block can corrupt future momentum selection/rebuild/priority behavior without re-verification. I’d store copies in the map and/or return `block.Copy()` from `BlockByHeight`.

**Verification**
Reviewed [[PR #43](https://github.com/zenon-network/go-zenon/pull/43)](https://github.com/zenon-network/go-zenon/pull/43), commit `e3697a1`.

Tests passed in a scratch copy of the PR:
- `go test ./chain/tests -run TestAccountPool -count=1`
- `go test ./chain/...`

0x3639:
@digitalSloth - Fable's review of copilot's comments.  Looks like we should try to address all three.  I will submit a PR against `dev` and ask for your review. 

---

I looked at Copilot’s 3 inline comments on PR #43. They’re all unresolved and not outdated.

1. `accountManager.Add` stores caller-owned block pointers  
I agree. This is the same root issue as my earlier finding, just from the insertion side. The pool now keeps mutable `*AccountBlock` pointers in `blocks`, so whoever still holds `transaction.Block` can mutate the pool after insertion. Better: store defensive copies in `Add`.

2. `accountManager.Pop` only deletes the frontier height  
Mostly agree, lower severity. For batched transactions with `DescendantBlocks`, `db.Pop()` rolls the frontier back before the whole batch, while `accountManager.Pop()` deletes only the main block height. Public range-based reads probably won’t return those stale descendant heights after rollback, but the map invariant is wrong and memory can leak. Better: record old frontier height, pop, record new frontier height, delete every cached height in `(newFrontier.Height, oldFrontier.Height]`.

3. `BlockByHeight` returns the internal pointer  
Strong agree. This is the clearest correctness risk and exactly what I flagged: before this PR, `ByHeight` deserialized a fresh block; now callers of `GetAllUncommittedAccountBlocks` can mutate pool state. Better: return a defensive copy.

My take: comments 1 and 3 should be fixed together; comment 2 is worth fixing while touching the same wrapper. One nuance: `AccountBlock.Copy()` does not deep-copy `PublicKey`, so if we want true defensive isolation we should either strengthen `Copy()` or add a local clone helper that copies `PublicKey` too.

PR #44, opened 2024-08-07
Improve account block validation performance

vilkris4:
Significant performance issues start appearing when the node validates an account block that uses an old momentum acknowledgement height.

The reason for the slow processing time of such an account block is that in order to verify the account block, the node must rollback the chain database's state to the block's momentum acknowledgement height. The rollback operation is done in the `Get` function in `versioned_db.go`.

In order to prevent the need for an expensive rollback operation, this PR introduces an additional chain cache database that stores the necessary state information in a way that can be accessed quickly.

To validate a *user* account block, the following stateful information is needed:

1. The amount of plasma fused to the sender's address at momentum acknowledgement height.
2. The amount of total chain plasma committed by the sender at momentum acknowledgement height.
3. What sporks are active at momentum acknowledgement height.

These three data items are stored in the cache using a momentum height composite key, allowing the key-value pairs to be iterated over in order to get the value at a specific height. Using the cache removes the need to rollback the momentum store to the momentum acknowledgment height, when validating a user account block.

To validate an *embedded* account block, various stateful data items are needed, since processing an embedded account block executes contract functionality. This means that the validation of an embedded account block still rolls back the momentum store to the momentum acknowledgement height. An embedded account block uses the send block's confirmation height as its acknowledgement height, which means that the amount of momentums needed to be rolled back will always be small, minimizing potential issues related to performance.

**Note:** An account block is not allowed to reference an older momentum acknowledgement height than what the previous account block has used. Because of this, the old states related to plasma are pruned. This significantly reduces the cache's size.

### Benchmarks
Benchmarking the time it takes to process the troublesome momentum 6,606,105 produces the following results:
|Version|Processing time|
|-----|-----|
|Current (v0.0.7)|10 min 21.6126 s|
|New|11.6853 ms|

### Testing

A full resync of the node has been tested with these changes.

### DIFF
--- a/chain/account/keys.go
+++ b/chain/account/keys.go
@@ -6,6 +6,9 @@ var (
 	chainPlasmaKey           = []byte{5}
 	receivedBlockPrefix      = []byte{6}
 	sequencerLastReceivedKey = []byte{7}
+
+	StorageKeyPrefix = storageKeyPrefix
+	ChainPlasmaKey   = chainPlasmaKey
 )
 
 const (
--- a/chain/cache.go
+++ b/chain/cache.go
@@ -1,0 +1,164 @@
+package chain
+
+import (
+	"fmt"
+	"sync"
+
+	"github.com/inconshreveable/log15"
+	"github.com/pkg/errors"
+	"github.com/zenon-network/go-zenon/chain/cache"
+	"github.com/zenon-network/go-zenon/chain/cache/storage"
+	"github.com/zenon-network/go-zenon/chain/nom"
+	"github.com/zenon-network/go-zenon/chain/store"
+	"github.com/zenon-network/go-zenon/common"
+	"github.com/zenon-network/go-zenon/common/db"
+	"github.com/zenon-network/go-zenon/common/types"
+)
+
+type chainCache struct {
+	manager storage.CacheManager
+	log     log15.Logger
+	changes sync.Mutex
+}
+
+func (c *chainCache) getFrontierStore() store.Cache {
+	if db := c.manager.DB(); db == nil {
+		return nil
+	} else {
+		return cache.NewCacheStore(storage.GetFrontierIdentifier(db), c.manager)
+	}
+}
+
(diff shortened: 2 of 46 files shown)


0x3639:
Here is a code review between George and Vilkris if you are interested in learning more about this PR.  

https://www.youtube.com/watch?v=Pb87qKrtszs&feature=youtu.be

digitalSloth:
@vilkris4 Thank you for this. Iv been reviewing and it looks good overall but Iv spotted a few potential issues - I'll note them here but I'll open a new PR to address them after merging this to `dev`

It looks like this refactor also changes a consensus rule, not just the lookup mechanism. In `verifier/account_block.go` (`fromHash`, ~line 350) and `vm/vm.go` (`applyReceive`, ~line 151), user account blocks now resolve their referenced send via `frontierStore` instead of the acknowledged-height momentum store. That means a receive block whose send confirmed after its own `MomentumAcknowledged` height (but before the current frontier) is now accepted, where the old rollback-based code would reject it. Since this changes what a block validates as valid, a pillar on this code could produce a momentum that old-code nodes reject.

Separately, `chain/tests/cache_test.go`'s `TestCache_Rollback` only rolls back empty momentums, so the pruning-then-rollback path (the riskiest part of this design — pruned entries need to come back correctly after a reorg) is never actually exercised. Worth adding a case that fuses/cancels plasma across the prune boundary, rolls back across it, and asserts the historical values are restored.

A few minor things noted inline/below, none blocking: fused-plasma entries for pure-beneficiary addresses are never pruned (keyed by sender, not beneficiary) so the "cache stays small" guarantee doesn't fully hold for those addresses; a couple of dead `ErrNotFound` branches in `chain/cache/account.go`; an unexplained `height==1` short-circuit in `chain/cache/spork.go`; and `NewCacheStore`/cache reads use `panic` rather than returning an error on bad input — not reachable on the current validated path but worth hardening given this is consensus code.

PR #45, opened 2024-08-09
Fix versioned db tests

vilkris4:
The test cases `TestVersionedDBConcurrentUse` and `TestVersionedDBVersions` in `versioned_db_test.go` will fail because the `LevelDBManager` instances are not cleaned up during teardown.

The following error messages will be shown when running the tests:

```TempDir RemoveAll cleanup: remove ...\TestVersionedDBConcurrentUse2208946391\001\000001.log: The process cannot access the file because it is being used by another process.```

```TempDir RemoveAll cleanup: remove ...\TestVersionedDBVersions80412342\001\000002.ldb: The process cannot access the file because it is being used by another process.```

This commit makes sure the resources are properly cleaned up during teardown.

### DIFF
--- a/common/db/versioned_db_test.go
+++ b/common/db/versioned_db_test.go
@@ -67,6 +67,7 @@ func newMockTransaction(seed int64, db DB) *mockTransaction {
 
 func TestVersionedDBConcurrentUse(t *testing.T) {
 	m := NewLevelDBManager(t.TempDir())
+	defer m.Stop()
 	v0 := m.Frontier()
 	v01 := m.Frontier()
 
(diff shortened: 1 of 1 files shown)


georgezgeorgez:
Ideally, I would want someone to recreate this issue. @vilkris4 mentioned this was on Windows.

openmac0815:
**Review — APPROVE**

Two added `Stop()` calls for leaked LevelDB managers in the versioned-DB tests. Checked the PR head: no double-`Stop` — the deferred stop in `TestVersionedDBConcurrentUse` is the only stop of that manager, and the second test stops `m` explicitly and `m2` via defer. Test-only change, safe to merge.

PR #47, opened 2024-11-25
governance v1

sumoshi21:
Implement the governance logic that is able to create and activate sporks and call administrative methods of other embedded contract.

### DIFF
--- a/common/types/address.go
+++ b/common/types/address.go
@@ -31,8 +31,9 @@ var (
 	AcceleratorContract = parseEmbedded("z1qxemdeddedxaccelerat0rxxxxxxxxxxp4tk22")
 	HtlcContract        = parseEmbedded("z1qxemdeddedxhtlcxxxxxxxxxxxxxxxxxygecvw")
 	BridgeContract      = parseEmbedded("z1qxemdeddedxdrydgexxxxxxxxxxxxxxxmqgr0d")
+	GovernanceContract  = parseEmbedded("z1qxemdeddedxg0vernancexxxxxxxxxxxklyh23")
 
-	EmbeddedContracts = []Address{PlasmaContract, PillarContract, TokenContract, SentinelContract, SwapContract, StakeContract, SporkContract, LiquidityContract, AcceleratorContract, HtlcContract, BridgeContract}
+	EmbeddedContracts = []Address{PlasmaContract, PillarContract, TokenContract, SentinelContract, SwapContract, StakeContract, SporkContract, LiquidityContract, AcceleratorContract, HtlcContract, BridgeContract, GovernanceContract}
 	EmbeddedWUpdate   = []Address{PillarContract, StakeContract, SentinelContract, LiquidityContract, AcceleratorContract}
 
 	SporkAddress *Address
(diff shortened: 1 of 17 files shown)


georgezgeorgez:
Hi,
I've created a set of maintainer criteria for myself to follow.
https://zenon.wiki/index.php/User:George

If you are looking for my approval on this PR, please answer the questions relevant to protocol changes.

0x3639:
@sumoshi21 do you want to fix the merge conflicts?

edgepillar:
Thanks for the governance proposal. I reviewed `3fef4c90505be733aad25e6350005708bacb7805` and found a quorum-accounting concern that needs an explicit electorate policy.

This is a source-level review; I have not executed a revoke/execute reproducer.

**[P1] Use the same electorate for stored votes and the acceptance threshold**

[`ExecuteAction`](https://github.com/zenon-network/go-zenon/blob/3fef4c90505be733aad25e6350005708bacb7805/vm/embedded/implementation/governance.go#L166-L213) uses the current number of active pillars for the threshold, while [`GetVoteBreakdown`](https://github.com/zenon-network/go-zenon/blob/3fef4c90505be733aad25e6350005708bacb7805/vm/embedded/definition/common.go#L577-L593) counts all stored votes without checking whether those pillars are still active. [Revocation](https://github.com/zenon-network/go-zenon/blob/3fef4c90505be733aad25e6350005708bacb7805/vm/embedded/implementation/pillars.go#L312-L353) updates the pillar record but does not remove its governance votes.

For a Type 1 action:

1. Ten active pillars cast six Yes and four No votes. The action fails because `600 <= 10 * 66`.
2. Two Yes voters revoke. The stored votes remain six Yes/four No, but the active-pillar count becomes eight.
3. The action now passes both checks: `6 > 4` and `600 > 8 * 66`.

An original-electorate snapshot would still have only 60% Yes; the current electorate would have four Yes/four No. The present result therefore does not match either interpretation.

This scenario requires owner-authorized revocations within their permitted windows, reflected in the consulted state, and execution before the proposal expires. It is not a claim of an unrestricted live exploit.

Could a maintainer confirm whether an action uses a fixed electorate or the current active electorate, then make both the vote tally and denominator follow that policy? If the hybrid behavior is intentional, its security rationale should be explicit. This should not silently change shared vote aggregation for other contracts.

A deterministic regression should record the six/four votes, revoke the two Yes voters, and attempt execution while the proposal remains valid. Please also cover No-voter revocation, new registrations during voting, and the exact acceptance boundary. The fixture should restore any global thresholds or spork state it changes.

I would resolve this policy/accounting issue before enabling governance authority over sporks and administrative methods.

openmac0815:
**Review — conditional approve (stale; needs rebase)**

The governance-v1 contract is properly spork-gated (GovernanceSpork in the `GetEmbeddedMethod` chain). One design point that deserves explicit release-note treatment: `ExecuteAction` can target the SporkContract, meaning pillar quorum gains the power to activate sporks — protocol-upgrade authority moves on-chain. That may be exactly the intent, but it should be stated plainly when this ships.

The 2024 base is well behind current `dev` (chain_bridge, downloader, RPC stack have all been reworked since). Needs a full rebase and a fresh pass before merge.

PR #58, opened 2026-05-19
feat: adds dockerised local devnet

digitalSloth:
Self-contained four-node Network of Momentum (NoM) for local development. 
Three pillars producing in rotation, plus a public RPC node. Chain ID `69`, fully isolated from mainnet.


digitalSloth:
Pointing this PR to the new `dev` branch and merging. 

All current feature branches depend on this work so we need to get it merged to avoid the noise in the other PRs

PR #62, opened 2026-06-16
docs(godoc): document every exported symbol across go-zenon

0x3639:
## Summary

A complete godoc pass over the entire go-zenon module. **203 `.go` files** gain
package- and symbol-level documentation, taking `golangci-lint run ./...` from a
large backlog of doc-lint findings to **0 issues over the whole repository** —
every exported symbol in go-zenon is now documented.

Every Go change is **comment-only** (no behavior change), proven per commit by
`scripts/check-comment-only.sh`. The branch additionally adds the lint/tooling
that enforces this going forward (`.golangci.yml`, the comment-only guard
script).

## What's documented

- **Common foundations** (`common`, `common/types`, `common/db`,
  `common/crypto`) — the storage and hashing layers, the core type surface, and
  the package-root utilities.
- **Chain / ledger** (`chain`, `chain/nom`, `chain/store`, `chain/genesis`) —
  the dual-ledger model (account blocks + momentums), the chain orchestration
  and stores, and genesis construction.
- **VM** (`vm`, `vm_context`, the abi layer, protocol constants) — the
  execution core, `ApplyBlock`/`ChangesHash` semantics, and the contract ABI
  layer.
- **Embedded contracts** (`vm/embedded/definition`,
  `vm/embedded/implementation`) — the on-chain data definitions and the
  execution flows for pillars, plasma, sentinel, token, stake, spork, swap,
  accelerator, htlc, liquidity and the bridge (wraps, unwraps, TSS and admin).
- **Consensus** (`consensus`, `consensus/storage`) — elections, points, tickers
  and events, and the consensus database.
- **Networking** (`protocol`, `p2p`) — the protocol sync/broadcast stack and the
  devp2p layer, including the syncer overview and the `minPeers` gate.
- **Node shell** (`wallet`, `pow`, `metadata`, `zenon`, `zenon/mock`, `node`,
  `verifier`, `pillar`, `app`, `cmd/znnd`, `rpc`, `rpc/server`) — HD-key wallet
  crypto, proof-of-work, build identity, the orchestrator lifecycle, the mock
  test harness, the consensus-critical verifier rule set, block production, the
  CLI, and the JSON-RPC 2.0 framework.

## Verification (held on every commit)

1. `GOWORK=off go build ./...` and the embedded-contract test suite pass.
2. `scripts/check-comment-only.sh master` — parse-level proof that every changed
   `.go` file is comment-only.
3. `GOWORK=off golangci-lint run ./...` — **0 issues** across the whole
   repository.

## Honesty over polish: bugs documented, not hidden

The godoc describes what the code **actually does**, not what it was meant to
do. Where the review surfaced a latent bug, the comment says so rather than
documenting intended-but-absent behavior. The notable ones are flagged inline,
e.g.:

- `wallet/manager.go` — `Lock` nils the map entry instead of deleting it, so
  `IsUnlocked` keeps reporting `true` after a lock (and `Stop` can panic on the
  nil entry); documented on `Lock`, `IsUnlocked` and `Stop`.
- `protocol/downloader/downloader.go` — `findAncestor`'s post-scan guard is
  inverted (`hash.IsZero()` vs `!hash.IsZero()`), making the binary-search
  fallback dead code and returning genesis on a deep fork; flagged in the
  `Downloader` doc and a `NOTE` on `findAncestor`.
- `app/cli.go` — `Stop`'s documented panic path was wrong (a nil `nodeManager`
  derefs before `DealWithErr`, which itself can never fire here); corrected.

No fixes are bundled here — this branch is documentation only.


0x3639:
I'll start by invoking the code review skill for this GitHub PR.

This is a large PR — 206 files, 9188 additions, claiming to be **comment-only** (documentation). The diff is too large for `gh pr diff`. Let me clone the repo locally to work with it properly.

The critical claim is **"every Go change is comment-only."** That's the user's #1 concern. I'll verify it independently rather than trusting the PR's own script. Let me write a robust AST-based check that strips comments and compares.

Now let me write an independent AST-based comment-stripping comparison tool.

My reprint-based tool is unreliable — `go/printer` uses source positions for line breaks, so removing comments can shift formatting and cause false "differs". Let me switch to the gold-standard method: **token stream comparison** via `go/scanner`, which ignores all whitespace and comments and compares only actual code tokens.

All 203 Go files are modifications (no added/deleted/renamed Go files). Now the definitive comment-only check:

Zero token differences reported. Let me confirm the loop actually executed all 203 comparisons (guarding against silent skips) with an explicit success counter.

**Confirmed: all 203 Go files have byte-identical token streams — no code was accidentally changed.** The comment-only claim holds independently.

Now let me review the 3 non-Go additions (the actual new "code"/config), then fan out to review comment correctness.

`.gitignore` adds a broad `docs/` pattern (ignores any `docs/` dir tree-wide) — minor. Now the important one, the guard script:

The guard script is sound (my token-based check is stronger and already authoritative for this PR). `.golangci.yml` and the script look correct.

Now the substantive task: verifying that **9,188 lines of new comments accurately describe the code**, across 203 files. This is where errors hide. The PR itself makes three strong "this is a latent bug" claims in comments that I must verify are accurate. I'll fan out parallel review agents by subsystem, then verify findings.

Let me set up tracking and dispatch.

All 203 Go files verified comment-only. Now dispatching parallel review agents to check comment **accuracy** against the actual code, subsystem by subsystem. The three PR-flagged "latent bug" comments get dedicated scrutiny.

Agents are reviewing in the background. While they work, let me cross-check the PR's own guard script against my independent token result, and confirm the tree still builds (the PR claims `GOWORK=off go build ./...` passes).

The build passes (only a third-party cgo deprecation warning), and the three flagged-bug comments are all **CONFIRMED-ACCURATE** against the code. Waiting for the remaining subsystem agents to report.

I've confirmed the two foundational facts already: **no code changed** (byte-identical token streams across all 203 files, cross-checked two independent ways) and the **three flagged-bug comments are all accurate**. Seven subsystem agents are still verifying comment accuracy across common/, chain/, vm core, vm/embedded, consensus/, rpc/, and node/verifier/net. I'll compile the full findings when they report back.

Consensus review complete — all accurate. Five subsystem agents still running (common/, chain/, vm core, vm/embedded, rpc/, node/verifier/net). Waiting for their reports.

common/ is clean too. Four agents remaining (chain/, vm core, vm/embedded, rpc/, node/verifier/net — five actually). Continuing to wait.

VM core reports two minor phrasing imprecisions (not factual errors) — noted for the summary. Still awaiting chain/, vm/embedded, rpc/, and node/verifier/net agents.

Node/verifier/net is clean with one minor wording nuance. Three agents remain: chain/, vm/embedded, and rpc/.

chain/ is clean with two minor looseness items. Two agents remain: vm/embedded and rpc/.

vm/embedded — the highest-risk area (bridge, token, stake formulas) — is fully accurate. Only the rpc/ agent remains.

All 8 agents are done. But the user asked to review **every** file — let me compute which of the 203 changed files were not covered by the agents' scopes and sweep the remainder.

73 files weren't fully covered. Dispatching a final sweep in four parallel agents:

Four sweep agents launched to cover the remaining 73 files. This achieves full coverage of all 203 changed Go files. Waiting for their reports to compile the final summary.

rpc leftovers are clean (three sub-threshold boundary nits noted). Three sweep agents remaining.

chain/common leftovers done — one minor over-claim in `common/ticker.go`. Two sweep agents remaining.

vm/abi + misc sweep found three minor inaccuracies worth listing. One agent left: p2p/protocol/pow/pillar.

All agents are done. Before finalizing, let me personally verify the two most significant findings (the p2p ones):

Both p2p findings confirmed first-hand. Here's the complete review.

---

# Peer Review: go-zenon PR #62 — `docs(godoc): document every exported symbol`

## 1. Does it accidentally change any code? — **No. Verified independently.**

I did not rely on the PR's own guard script. I checked out the branch and built my own verifier using Go's `go/scanner`: it tokenizes each file (comments dropped by the scanner itself) and compares the exact token streams of base vs. PR head.

- **All 203 changed `.go` files: token streams byte-identical** (`identical=203, differ=0, scan_err=0`). No added/deleted/renamed Go files.
- Cross-checks: the PR's own `scripts/check-comment-only.sh master` also passes, and `GOWORK=off go build ./...` succeeds on the branch (only a pre-existing third-party cgo deprecation warning).

So every Go change is provably comment-only. The only functional additions are three non-Go files:

- **`.golangci.yml`** — correct golangci-lint v2 config; enables only `revive` `exported` + `package-comments`. Sound.
- **`scripts/check-comment-only.sh`** — I audited it line by line: parse-both-sides/strip-comments/diff approach is correct, it properly rejects new/deleted files and separately pins build directives (`//go:`, `+build`) so those can't be smuggled in as "comments." Sound.
- **`.gitignore`** — adds `docs/`. ⚠️ Note: this is an **unanchored** pattern, so it ignores *any* directory named `docs` anywhere in the tree, and it will hide a future intentional `docs/` from `git status`. Consider `/docs/` and/or a follow-up to remove it once the internal working tree is gone.

## 2. Are the comments correct? — Reviewed all 203 files, 9,188 added lines

I fanned out 12 review agents by subsystem with explicit file lists covering every changed file, each diffing the added comments and reading the real code (and callers) behind every checkable claim — constants, formulas, key layouts, error semantics, crypto parameters, ordering rules. I then re-verified the top findings myself.

**The PR's three self-declared "latent bug" comments are all accurate** (verified against the code, not the comments):
- `wallet/manager.go` — `Lock` really does `m.decrypted[path] = nil` instead of `delete`; `IsUnlocked` uses comma-ok so it stays `true`; `Stop` calls pointer-receiver `Zero()` on the nil entry → real panic path. All three sub-claims true.
- `protocol/downloader/downloader.go:458` — the `findAncestor` guard really is inverted (`hash.IsZero()` where `!hash.IsZero()` is intended): no-match returns genesis, match falls into a redundant binary search. NOTE is accurate.
- `app/cli.go:46` — nil `nodeManager` deref happens before `DealWithErr`, and both `Manager.Stop()` variants always return nil, so `DealWithErr` can never fire. Corrected comment is accurate.

Two more latent-bug comments (`FusionEntryList.UnmarshalJSON` in `rpc/api/embedded/plasma.go`, `Project.UnmarshalJSON` in accelerator.go — slice-sizing panics) were also verified as real, not fabricated.

The heavy numeric/behavioral claims all check out: plasma constants (21000 base, 2100/fusion unit, 1500 difficulty/plasma), stake weight `(9+units)/10`, pillar/sentinel lock windows, bridge wrap/unwrap ordering and TSS delay distinctions, BIP-44 path `m/44'/73404'/index'` + SLIP-0010 + Argon2id + AES-256-GCM, election two-ticks-back proof momentum, the account-block-hash-excludes-ChangesHash vs momentum-hash-includes-it asymmetry, verifier rules, DB key prefixes, genesis timestamp, RPC pagination limits and error-code ownership.

## 3. Issues found

**Factual inaccuracies worth fixing before merge (4):**

1. **`p2p/discover/table.go:51-52`** — new `Table` doc says buckets keep nodes "most recently active **last**." Inverted: `bucket.bump()` moves the active node to `entries[0]` ("move it to the front") and eviction treats `entries[bucketSize-1]` as *oldest*. Most-recently-active is **first**. (It repeats a pre-existing wrong comment on `bucket`, but new docs shouldn't restate it.)
2. **`p2p/peer_error.go:54-56`** — claims the `DiscReason` "numeric values are fixed by the devp2p wire protocol." The plain-`iota` block makes `DiscSubprotocolError` = 12 (0x0c), while devp2p/go-ethereum fix it at **0x10**. True for 0x00–0x0b only.
3. **`vm/abi/type.go:12-14`** — says "AddressTy, TokenStandardTy and HashTy are Zenon-specific." `AddressTy` and `HashTy` exist in upstream go-ethereum's ABI; only `TokenStandardTy` is Zenon-added.
4. **`vm/vm_context/account_context.go:24-25`** — `MomentumStore()` doc says the store is "fixed at the block's acknowledged momentum." Only the supervisor path does that; `rpc/api/utils.go` passes the frontier store and genesis contexts pass `nil` (as the constructor's own doc admits three lines below).

**Minor wording imprecisions (optional polish, none false in substance):**

- `vm/vm.go` (~80): plasma "capped at MaxPlasmaForAccountBlock" — the code *rejects* (`ErrBlockPlasmaLimitReached`), it doesn't clamp.
- `vm/supervisor.go` (~46): failed calls rolled back "with a refund" — no refund block when the send carried zero tokens.
- `pow/pow.go:99`: digest must "exceed" threshold — comparison accepts equality (`≥`).
- `protocol/broadcaster.go:19-21`: fetcher runs "once the node has caught up" — only the full-momentum path is sync-gated; hash announcements aren't.
- `consensus/chain_ticker.go:23`: `GetContent` excludes the genesis momentum at tick 0 despite it falling in the tick.
- `common/ticker.go:16`: "momentum, epoch and point schedules as tickers" — no per-momentum ticker exists; only election-tick and epoch.
- `zenon/zenon.go:1`: protocol's "Init is a no-op" — the call is actually commented out, not a called no-op.
- `chain/genesis/mock/embedded_genesis.go` package doc: "the pillar" understates funding of Pillar1–Pillar8.
- `chain/interface.go`: dangling cross-reference to a fork-priority rule the rewritten doc no longer states (rule still documented on `higherPriority`).
- `rpc/api/embedded/bridge.go:54, 200` and `spork.go:16`: unhalt boundary is off-by-one vs "have passed"; "between heights X and Y" leaves inclusivity ambiguous; `Count` is parseable networks, not all registered.

## Verdict

**Safe to merge from a code-safety standpoint** — the comment-only claim is independently proven at the token level, and the tooling additions are sound. The documentation is unusually rigorous and honest (the flagged latent bugs are all real). I'd request the 4 factual fixes above (items 1–4) — all one-line comment edits — and treat the minor list as take-it-or-leave-it polish.

0x3639:
Addressing merge conflicts now that we are on the dev branch.

edgepillar:
Reviewed at `7997c56`. This is a source-only review; I have not independently executed the guard, tests, builds or reported lint validation.

Two points before merging into `dev`:

1. **Reconcile the documentation with the target implementation.** The reviewed `dev` snapshot at `32b96d9` already uses `!hash.IsZero()` in `findAncestor`, while this PR's new documentation describes the inverted guard and an unreachable binary-search fallback. Please update those claims when reconciling the branch, preserving the existing fix. The legacy/libp2p reorganization also needs consideration; verification against `master` does not establish the accuracy or safety of the eventual target-tree integration. [PR documentation](https://github.com/zenon-network/go-zenon/blob/7997c5604ae505e5291d80a0bb1a45ee7954658f/protocol/downloader/downloader.go#L400-L407), [dev implementation](https://github.com/zenon-network/go-zenon/blob/32b96d9241a53966c31c310bd637e57562300255/protocol/downloader/downloader.go#L425-L431).

2. **Qualify the comment-only guard's guarantee.** The stripped-source comparison and separate directive-text comparison are useful, but unchanged directive text can move between declarations, and cgo preambles can carry semantics despite being comments. Please either cover those cases with negative fixtures or narrow the stated guarantee. I am not claiming this patch contains such a change. [Guard](https://github.com/zenon-network/go-zenon/blob/7997c5604ae505e5291d80a0bb1a45ee7954658f/scripts/check-comment-only.sh#L43-L78).

The complete diff's lexical inspection found no non-comment token differences across the 203 modified Go files, but that is not a compiled or conflict-resolved integration proof.

openmac0815:
**Review — approve in principle**

Spot-checked: changes are doc comments on exported symbols plus a revive lint config (`exported`, `package-comments`) that makes the documentation enforceable in CI — good combination. Caveat: the diff exceeds GitHub's 20k-line API limit, so I could not verify every file mechanically; and a tree-wide docs PR bit-rots quickly against the active hardening series (#81–#103 touch many of the same files). Recommend merging soon or regenerating from a fresh base to avoid a long conflict tail.

PR #67, opened 2026-07-10
fix: restore acknowledged-height scoping for user receive validation

digitalSloth:
verifier/account_block.go: fromHash() resolved the referenced send via frontierStore, which only checks existence at the current tip. Add a GetBlockConfirmationHeight comparison so a receive is rejected when its send confirmed after the receive's own MomentumAcknowledged height, matching the rollback-based validation this replaces. Adds a regression test pinning the boundary.

chain/tests/cache_test.go: add TestCache_RollbackAfterPrune, exercising the prune-then-rollback-restore path (fuse/cancel across the prune boundary, roll back, assert restoration, re-apply forward for idempotency) that existing tests didn't cover.

Minor hardening in the cache package: document why fused-plasma entries for pure-beneficiary addresses aren't pruned, return nil instead of panicking when NewCacheStore's height is out of range, propagate enoughPlasma's errors instead of DealWithErr, drop dead ErrNotFound branches, remove an unexplained height==1 short-circuit in spork lookup, and note the merged DB iterator's reverse-iteration limitation.

### DIFF
--- a/chain/cache/account.go
+++ b/chain/cache/account.go
@@ -3,7 +3,6 @@ package cache
 import (
 	"math/big"
 
-	"github.com/syndtr/goleveldb/leveldb"
 	"github.com/zenon-network/go-zenon/chain/nom"
 	"github.com/zenon-network/go-zenon/common"
 	"github.com/zenon-network/go-zenon/common/types"
@@ -24,9 +23,6 @@ func getChainPlasmaKeyPrefix(address []byte) []byte {
 
 func (cs *cacheStore) GetStakeBeneficialAmount(address types.Address) (*big.Int, error) {
 	value, err := cs.findValue(getFusedAmountKeyPrefix(address.Bytes()))
-	if err == leveldb.ErrNotFound {
-		return big.NewInt(0), nil
-	}
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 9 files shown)


0x3639:
Codex 5.6 sol

Request changes. The fresh review found one merge-blocking issue missed previously, plus the two existing test-quality findings.

1. **[P1] Height-1 spork behavior now diverges from the canonical ledger.**  
   [chain/cache/spork.go:17](/tmp/go-zenon-pr67.SbGSyG/chain/cache/spork.go:17) removes the genesis-height guard, while [chain/momentum/embedded.go:139](/tmp/go-zenon-pr67.SbGSyG/chain/momentum/embedded.go:139) still requires every spork to be inactive at height 1. Custom genesis files may legally contain activated sporks, so blocks acknowledging genesis would use different VM rules after this PR. I reproduced the cache returning `true` for that case. Restore the guard unless changing genesis consensus behavior is intentional and coordinated.

2. **[P3] Restore the mutated global.**  
   [chain/tests/cache_test.go:186](/tmp/go-zenon-pr67.SbGSyG/chain/tests/cache_test.go:186) leaves `constants.FuseExpiration` set to `5`. Save the original value and restore it with `t.Cleanup`.

3. **[P3] Match the production rollback boundary.**  
   [chain/tests/cache_test.go:251](/tmp/go-zenon-pr67.SbGSyG/chain/tests/cache_test.go:251) uses `<`, but production permits a rollback distance equal to `GetRollbackCacheSize()`. Use `<=`.

The acknowledged-height receive fix itself is correct: confirmation heights are ledger-versioned and the comparison restores the old snapshot-scoped validation. The cache rollback regression test is also valuable once its isolation issues are fixed.

Local validation passed for `./chain/...`, `./verifier`, `./vm/...`, and `./rpc/api/...`; `git diff --check` passed. GitHub currently reports no configured check runs for head `ddc499b`.

digitalSloth:
Thanks for the review

Height-1 spork guard (@0x3639): good catch. The cache does hold genesis spork state (Init replays the height-1 patch through the extractor, and genesis config writes activated sporks there), and the VM reads spork status via the cache in vm/vm_context/spork.go. So a custom genesis with an activated spork at height 1 would diverge from momentumStore.IsSporkActive, which still forces false at height 1. Restored the guard so the cache mirrors the canonical ledger.

FuseExpiration not restored: fixed — now saved and restored via t.Cleanup.

Rollback boundary < vs <=: fixed — changed to <= to match RollbackCacheTo, which only errors when the distance exceeds GetRollbackCacheSize().

0x3639 (approved):
No remaining code findings. All three issues are correctly addressed on head `bbfe5e9`, and the full relevant suite passed.  

thx!

PR #56, opened 2026-04-28
chore(deps): bump golang.org/x/crypto v0.1.0 -> v0.31.0 (CVE-2024-45337)

mehowz:
## Summary

Bumps `golang.org/x/crypto` from `v0.1.0` (Oct 2022) → `v0.31.0` (Dec 2024) to close the dep-audit signal for [CVE-2024-45337](https://nvd.nist.gov/vuln/detail/CVE-2024-45337).

## Re-scoped from #54

This PR is the corrected version of the now-closed [#54](https://github.com/zenon-network/go-zenon/pull/54). The previous PR was branched from a fork branch that had unrelated divergence (new embedded-contract definitions from a separate work branch — `2b8232c`), which Copilot + maintainer review correctly flagged as out-of-scope. This branch is cut directly from `upstream/master` so the scope is strictly the dep bump.

**Diff: 2 files changed, 9 insertions, 13 deletions** (just `go.mod` + `go.sum`).

## Exploitability

The vulnerable `golang.org/x/crypto/ssh` subpackage is **not imported in this repo** — `sha3`, `ripemd160`, and `argon2` are the only sub-packages used. The CVE is therefore not directly exploitable here. This PR is dep-graph hygiene.

```
$ git grep -l "x/crypto/ssh" -- '*.go'
(empty)
```

## Side effect

- `golang.org/x/sys`: v0.1.0 → v0.28.0 (transitive)

## Verification

- `go build ./...` clean (one unrelated gopsutil/IOKit deprecation warning, pre-existing on master)
- `go test -short ./...` — all packages pass

Apologies for the noise on #54 — that was process error on my side, not a deliberate scope expansion. Verified my workflow now branches from `upstream/master` for upstream PRs.

PR #52, opened 2025-11-01
Fix isRevocable field in plasma getEntriesByAddress API

0x3639:
The getEntriesByAddress method now correctly returns the isRevocable boolean field for fusion entries based on whether the current momentum height has reached the entry's expiration height.

Changes:
- Added IsRevocable field to FusionEntry and FusionEntryMarshal structs
- Updated JSON marshaling methods to handle IsRevocable
- Modified GetEntriesByAddress to calculate isRevocable status
- Updated all plasma tests to reflect new field in responses

Issue #48, opened 2024-12-12
RPC Unmarshaling Bug


georgezgeorgez:
There are several rpc unmarshalling bugs.
The pattern seems to be trying to pre-allocate the slice but using its own length instead of the auxiliary data.
It seems like a copy and paste from the marshalling functions that were not replaced fully.

Example:
```
aux.Phases = make([]*Phase, len(p.Phases))
p.Phases = make([]*Phase, len(p.Phases))
```

I will first gather all the instances I can find and them make an MR if no one finds others.

https://github.com/zenon-network/go-zenon/blob/429e853ba038be1eddc5c0708cb26cff79610101/rpc/api/embedded/accelerator.go#L152

https://github.com/zenon-network/go-zenon/blob/429e853ba038be1eddc5c0708cb26cff79610101/rpc/api/embedded/plasma.go#L149

georgezgeorgez:
TODO
should use copy(to, from) instead of a loop (S1001)go-staticcheck

PR #68, opened 2026-07-13
Bugs & Performance

digitalSloth:
# Bugs & Performance

This PR promotes the accumulated `dev` work to `testnet` ahead of spinning up a new testnet. It bundles ten merged PRs spanning node performance, p2p/consensus safety, RPC correctness, and a security dependency bump.

## Highlights

- **Two major performance reworks** to the account pool and account-block validation, together turning multi-minute worst cases into single-digit milliseconds.
- **A p2p safety fix** that stops momentums being broadcast before local insertion and closes a remotely-triggerable nil-pointer crash.
- **Four RPC/API correctness fixes** affecting pagination, plasma, unwrap ordering, and JSON unmarshalling.
- **A security dependency bump** clearing CVE-2024-45337.

---

### Account pool performance (#43, hardened by #64)

The account pool degraded badly when holding many pending blocks across account chains with long unconfirmed states — `getUncommittedAccountBlocksByAddress` and `rebuild` were doing repeated reads against the unconfirmed (frontier) DB. A new in-memory `accountManager` keeps each account's pending blocks in a `map[uint64]*nom.AccountBlock`, removing those reads.

Benchmark (`GetAllUncommittedAccountBlocks`, 2,500 pending blocks across five accounts): **4,477 ms/op → 4.33 ms/op** (~1000×).

**#64** hardens the cache after review found it shared mutable state with callers: `Add` now stores a deep copy and indexes the copy's descendants; `Pop` deletes the full rolled-back height range instead of just the frontier height; `BlockByHeight` returns a defensive copy. A supporting fix makes `nom.AccountBlock.Copy()` also copy the `PublicKey` field. Defensive copies cost ~12% (4.87 ms/op), preserving the ~1000× win. Adds unit and integration tests for the pop-range eviction and copy-on-read/write isolation.

### Account block validation performance (#44, corrected by #67)

Validating an account block that references an old momentum acknowledgement height forced a rollback of the chain DB state to that height (in `versioned_db.go`) — pathologically slow. A new chain cache DB stores the state needed to validate a *user* account block (fused plasma at ack height, total committed chain plasma at ack height, active sporks at ack height) keyed by momentum height, eliminating the rollback. Embedded account blocks still roll back, but only to the send block's confirmation height, which is always shallow. Old plasma states are pruned since a block can't reference an older ack height than its predecessor.

Benchmark (processing momentum 6,606,105): **10 min 21.6 s → 11.69 ms**.

**#67** is the correctness follow-up to the cache work. `verifier/account_block.go`'s `fromHash()` resolved the referenced send via the frontier store, which only checks existence at the current tip; it now compares against `GetBlockConfirmationHeight` so a receive is rejected when its send confirmed *after* the receive's own acknowledged height — restoring the scoping the rollback-based path enforced. Adds a boundary regression test and `TestCache_RollbackAfterPrune` covering the prune-then-rollback-restore path, plus cache-package hardening (return nil instead of panicking on out-of-range height in `NewCacheStore`, propagate `enoughPlasma` errors, remove dead `ErrNotFound`/`height==1` branches, document pruning and iterator limitations).

### Momentum verification before broadcast (#60)

The fetcher used to broadcast a fetched momentum to peers *before* `InsertChain`. If insertion then failed (invalid content, VM error, fork), the momentum had already been announced — producing "ghost hashes" that `getFrontierMomentum` could return but that never appeared over RPC. The `validateBlock` callback meant to gate this was a no-op.

`insert()` now runs: parent-known check → `verifyBlock` (cheap structural pre-check via `VerifyMomentum`: chain id, version, timestamp, previous-hash linkage, content ordering — drops the peer on failure) → `insertChain` (authoritative verification incl. signature/producer checks; aborts with no broadcast on failure) → broadcast. A momentum is only relayed after a successful local insert. Also fixes a nil-pointer panic in `rawMomentumVerifier.content()` — it called `isBatched(block)` before checking the account-block lookup succeeded, so any peer could crash the node with a mismatched content header and no valid signature. Removes the dead `validator` closure in `protocol/handler.go`. Adds fetcher and verifier regression tests.

## RPC / API correctness

- **#51 — `GetAccountBlocksByPage` `more` flag.** The `more` property always returned false, so clients couldn't tell there were more pages. Now set based on `startHeight` (older blocks remain when `startHeight > 1`).
- **#52 — plasma `getEntriesByAddress` `isRevocable`.** Adds an `IsRevocable` field to `FusionEntry`/`FusionEntryMarshal` and their JSON marshalling, computed from whether the current momentum height has reached the entry's expiration height. Plasma tests updated.
- **#57 — `GetAllUnwrapTokenRequests` ordering.** Now sorts by `RegistrationMomentumHeight` descending (newest first) to match the wrap API, giving consistent ordering across wrap/unwrap endpoints.
- **#49 — RPC `unmarshalJSON` bugs.** Fixes the JSON unmarshalling bugs tracked in #48.

## Dependencies / security

- **#56 — bump `golang.org/x/crypto` v0.1.0 → v0.31.0 (CVE-2024-45337).** Dep-graph hygiene: the vulnerable `x/crypto/ssh` subpackage isn't imported here (only `sha3`, `ripemd160`, `argon2`), so the CVE isn't directly exploitable, but the bump clears the audit signal. Pulls `golang.org/x/sys` v0.1.0 → v0.28.0 transitively. `go build ./...` and `go test -short ./...` pass.

---

## Included PRs (in merge order)

| # | Title | Area |
|---|-------|------|
| #51 | Fix `GetAccountBlocksByPage` to properly set `more` property | RPC |
| #52 | Fix `isRevocable` field in plasma `getEntriesByAddress` API | RPC |
| #57 | Fix `GetAllUnwrapTokenRequests` to sort by `RegistrationMomentumHeight` | RPC |
| #43 | Improve account pool performance | Performance |
| #64 | Defensively copy cached blocks (addresses #43 review) | Performance |
| #44 | Improve account block validation performance | Performance |
| #67 | Restore acknowledged-height scoping for user receive validation | Performance / correctness |
| #60 | Enable momentum verification before broadcast | p2p / consensus |
| #49 | Fix RPC `unmarshalJSON` bugs | RPC |
| #56 | Bump `golang.org/x/crypto` (CVE-2024-45337) | Security |

## Testing notes

Full node resyncs were tested for both #43 and #44. New unit/integration/regression tests were added for the account-pool cache (#64), the chain cache rollback/prune path and receive-validation boundary (#67), and the fetcher/verifier broadcast ordering and panic fix (#60). Recommend a fresh resync plus a full `go test ./...` on the merged branch before cutting the testnet genesis.


0x3639:
Verdict: request changes on PR #68. I found two cache-consistency blockers and one resource leak.

- **[P1] Cache and canonical-chain commits are not atomic.** Both [[remote insertion](https://github.com/zenon-network/go-zenon/blob/e01f5c2560c1a9af0f30ea560e09fb5205077772/protocol/chain_bridge.go#L215-L221)](https://github.com/zenon-network/go-zenon/blob/e01f5c2560c1a9af0f30ea560e09fb5205077772/protocol/chain_bridge.go#L215-L221) and [[local momentum creation](https://github.com/zenon-network/go-zenon/blob/e01f5c2560c1a9af0f30ea560e09fb5205077772/protocol/broadcaster.go#L34-L40)](https://github.com/zenon-network/go-zenon/blob/e01f5c2560c1a9af0f30ea560e09fb5205077772/protocol/broadcaster.go#L34-L40) advance the cache before adding the momentum. If `AddMomentumTransaction` fails, the cache remains one momentum ahead. A later insertion can then apply against contaminated state and overwrite the same-height rollback record. Roll the cache back on failure or combine both transitions transactionally.

- **[P1] `ApplyMomentum` errors are discarded.** In [[chain/cache.go](https://github.com/zenon-network/go-zenon/blob/e01f5c2560c1a9af0f30ea560e09fb5205077772/chain/cache.go#L62-L68)](https://github.com/zenon-network/go-zenon/blob/e01f5c2560c1a9af0f30ea560e09fb5205077772/chain/cache.go#L62-L68), execution continues into `Changes` and `manager.Add` even if extraction or pruning failed. That can commit a partial cache patch while advancing its frontier. This matches the existing Copilot thread and should block merging.

- **[P2] Rollbacks leak LevelDB snapshots.** [[getRollback](https://github.com/zenon-network/go-zenon/blob/e01f5c2560c1a9af0f30ea560e09fb5205077772/chain/cache/storage/db.go#L149-L160)](https://github.com/zenon-network/go-zenon/blob/e01f5c2560c1a9af0f30ea560e09fb5205077772/chain/cache/storage/db.go#L149-L160) ignores `GetSnapshot` errors and never calls `Release`. Repeated reorgs retain snapshots and prevent obsolete LevelDB state from being reclaimed.

Validation on unchanged head `e01f5c2560c1`:

- `go test -short ./...` passed.
- Targeted race tests passed.
- Docker Compose validation and `go mod tidy -diff` passed.
- `go vet` stops on a pre-existing unreachable-code warning in `chain/account/mailbox/mailbox.go`.

digitalSloth:
I have open a PR addressing the points raised by both CoPilot and @0x3639 

https://github.com/zenon-network/go-zenon/pull/69

If there is any other feedback I'll address in that PR to

PR #70, opened 2026-07-21
ci: add PR test workflow, remove flaky UPnP discovery test

digitalSloth:
Adds a GitHub Actions workflow that builds and runs the full test suite on pull requests targeting dev, testnet, and master.

Removes p2p/legacy/nat/natupnp_test.go: TestUPNP_DDWRT hangs waiting on a multicast UDP response in sandboxed/CI network environments with no timeout, and nothing else in the package depends on its fakeIGD helper.


0x3639 (approved):
@digitalSloth wanted to bring your attention to > Notable finding — the deadlock is in production code.  Make sure to read that below.  Maybe we should submit a separate PR to fix that.  

# Review of zenon-network/go-zenon#70

**Verdict: Approve with minor suggestions.** Small, focused PR (+34/−244): a new PR-triggered test workflow plus deletion of one network-dependent test. I verified the action pins, the deleted test's isolation, and the hang claim against the actual discovery code — the diagnosis holds up, and it surfaces a real production bug worth a follow-up.

## What I verified

- **All three action SHAs resolve to real upstream commits**: `actions/checkout@fbc6f399` = v5.1.0, `actions/setup-go@d35c59ab` = v5.5.0, and the `ensure-sha-pinned-actions` pin is byte-identical to the one already used in `znn_builder.yml`. No supply-chain concerns.
- **The gate job won't break existing CI**: `ensure-sha-pinned-actions` scans every workflow in the repo, but `znn_builder.yml` is already fully SHA-pinned, so nothing regresses.
- **`go-version-file: go.mod`** resolves to Go 1.23 on the `dev` base branch — correct.
- **`fakeIGD` has no other users**: only `natupnp_test.go` defined or referenced it; `nat.go`, `nat_test.go`, and `natpmp.go` are untouched by the deletion.
- **The hang claim is accurate, and the test could never pass in Actions anyway.** In `p2p/legacy/nat/natupnp.go`, `discover()` returns *without sending to the channel* when `goupnp.DiscoverDevices` errors, so `discoverUPnP()` blocks forever on `<-found` — that's the no-timeout hang. And even when discovery merely returns empty, the deleted test called `t.Fatalf` whenever `CI` is set (GitHub Actions always sets it), so removal (or an always-skip) was required for this workflow to be green.

## Notable finding — the deadlock is in production code, not just the test

Deleting the test removes the symptom, but `discover()`'s early `return` on error still deadlocks `discoverUPnP()`, which is live code: `nat.go:155` and `nat.go:169` (`startautodisc("UPnP", discoverUPnP)`) call it for nodes configured with UPnP NAT. In an environment where the SSDP multicast send errors, that goroutine blocks forever and NAT resolution never completes. Upstream go-ethereum fixed this with a one-liner — send `nil` before returning on error:

```go
func discover(out chan<- *upnp, target string, matcher func(*goupnp.RootDevice, goupnp.ServiceClient) *upnp) {
	devs, err := goupnp.DiscoverDevices(target)
	if err != nil {
		out <- nil // was: bare return, leaving discoverUPnP blocked forever
		return
	}
```

Non-blocking for this PR, but worth filing as a follow-up or adding the one-line fix here.

## Workflow suggestions (all non-blocking)

- **Add `timeout-minutes` to the `test` job** (e.g. 30). Without it, a hung test burns the 6-hour default runner limit — ironic given this PR exists because of a hanging test. Go's per-package 10-minute test timeout softens this, but a job-level cap is cheap insurance.
- **Add `permissions: contents: read`** at the workflow level. The default token grant is broader; explicitly minimizing it fits the hardening theme the `harden_security` job establishes.
- **Add a `concurrency` group** with `cancel-in-progress: true` keyed on the PR ref, so force-pushes don't stack redundant runs of a serial, single-threaded test suite.
- **Test loop**: the per-package `for` loop with `-failfast` gives stop-at-first-failing-package, which a single `go test -p 1 ./...` wouldn't — reasonable trade-off, just note it re-runs `go vet` per invocation and forfeits some scheduling. `-p 1` is effectively redundant inside a one-package invocation but harmless. Fine as is.
- **Consistency nit**: this workflow pins checkout v5.1.0 while `znn_builder.yml` pins v4.2.2. Works fine (ubuntu-latest supports node24), but aligning versions — and optionally adding `# vX.Y.Z` comments beside the SHAs for Dependabot/readability — would help future maintenance.

## Test coverage

Net coverage loss is minimal: the deleted test exercised vendored go-ethereum discovery parsing against a fake device, was environment-dependent by design, and was structurally unable to pass in CI. The rest of the `nat` package's tests (`nat_test.go`) are unaffected.

PR #72, opened 2026-07-21
feat: multisig

digitalSloth:


lordBizness:
# Consolidated review — PR #72 `feat: multisig` (zenon-network/go-zenon)

*Reviewers: Claude (primary) + Codex (independent second pass over the same diff and checkout). All findings below were cross-verified by both reviewers; agreement level noted where they differed.*

## Overview

Adds protocol-level **mutable multisig accounts**: a new address class (prefix byte `2`, derived from `(creatorPubKey, nonce)` so signer rotation never changes the address), a spork-gated registry embedded contract (`CreateMultisig` / `ChangePolicy`) with a 60-momentum policy-maturity delay, threshold-signature verification in the account verifier (`MultisigAuth` carried outside the block hash, mirroring the existing `Signature` field), a `MomentumAcknowledged` recency floor (`MultisigMaxMaLag = 60`) enforced authoritatively at momentum-content verification, a producer-side filter that drops stale "poison" blocks to protect momentum-production liveness, a read-only RPC module, and wallet helpers. 34 files, +4420/−70, with tests at every level (codec, verifier-struct, contract-method, full-chain conformance including reorg, replay, race, and liveness scenarios).

**Verdict: high-quality, security-conscious work. No consensus-breaking bug found by either reviewer.** Items 1–3 below are worth addressing or explicitly acknowledging on the PR before merge; 4–7 are cheap hardening; the rest are nits.

## Verified sound (both reviewers)

- **Wire compatibility**: `rlp:"optional"` on the last field with old↔new interop tests in both directions; protobuf field 24 appended; JSON round-trip covered.
- **Deterministic spork gating**: activation is checked via the momentum store at the block's `MomentumAcknowledged`, not the node-local frontier.
- **Recency floor layering is correct**: the node-local pre-filter (`verifier/account_block.go` `momentumAcknowledged()`) cannot reject valid historical blocks during sync or ≤30-deep reorg import — `chain_bridge` applies blocks while the frontier sits at the containing momentum's parent, so `MA+60 ≥ momentumHeight > frontierH` always holds for content()-valid blocks. The consensus-authoritative floor lives in `rawMomentumVerifier.content()`, anchored to the including momentum's own height.
- **Liveness fix**: the producer filter drops stale multisig blocks contiguously per account (stale parent ⇒ descendants dropped too), preventing the orphaned-child momentum halt; regression-tested.
- **Policy state machine**: `Promote()` is the single source of truth for effective policy; matured pending is materialized before staging (no reversion across maturity boundaries, proven at unit and full-chain level); lock is monotonic once active; `ChangePolicy` receives are re-authorized against the post-maturity policy, closing the backdated-MA hole for policy mutations.
- **Funds safety on failure**: a failed `CreateMultisig` receive refunds the 1 ZNN burn via the existing `rollbackEmbedded` path.
- **Signature verification**: single canonicalization codec (sort/dedup/size) on the write path; duplicate-signature trial-match cannot double-count a signer; junk/truncated/non-member signatures rejected; verify loop bounded at 16×16 ed25519 ops.

## Findings

### Should address before merge

**1. Residual spend authority after rotation (design tradeoff — document prominently).** *[both AGREE]*
The verifier validates a block against the policy active at its (backdatable) `MomentumAcknowledged` height, and inclusion allows up to `MultisigMaxMaLag = 60` momentums of lag (`verifier/account_block.go:476`, `verifier/momentum.go:198`). So a superseded (possibly compromised) signer set retains authority over **plain value-transfer sends** for up to ~60 momentums (~10 min) after a rotation matures. Only `ChangePolicy` gets receive-time re-authorization (`vm/embedded/implementation/multisig.go:187`). Total exposure after initiating a key rotation ≈ maturity delay (60) + residual window (60) ≈ 20 min. The `MultisigMaxMaLag` comment acknowledges this, but it should be stated in user-facing docs so operators understand rotation does not immediately revoke a leaked key.

**2. Stale dormancy comment in `common/types/spork.go:39`.** *[both AGREE]*
The `MultisigSpork` comment claims the `embedded.go`/`vm_context` wiring "lands later; until then … IsSporkActive always returns false" — but this PR includes that wiring (`vm/embedded/embedded.go:257`, `vm/vm_context/spork.go:32`) and the devnet genesis activates the spork. The dormancy claim is now false as written; update the comment.

**3. Stored records are trusted without decode-time validation.** *[Codex, endorsed by Claude — supersedes the narrower threshold-0 point]*
`ParseMultisigRecord` (`vm/embedded/definition/multisig.go:215`) reconstructs policies without calling `ValidPolicy` — no check of threshold bounds, signer count, uniqueness, or canonical ordering. Separately, `VerifyThresholdSignatures` (`multisig.go:278`) vacuously passes with `Threshold == 0` and zero signatures. Neither is reachable today (every write path validates), but this is a consensus-critical read path: a future codec/migration bug could activate the threshold-zero bypass. Recommend (a) `if policy.Threshold < 2 { return false }` in `VerifyThresholdSignatures`, and/or (b) validating decoded records in `ParseMultisigRecord`.

### Hardening (nice to have)

**4. No decode-time cap on `MultisigAuth` size.** *[both AGREE]*
Signature count is only checked against the threshold inside `signature()`; sig lengths only inside `VerifyThresholdSignatures`. The p2p transport's 10 MiB message ceiling still permits far more decode/allocation than the protocol-valid maximum (~1 KiB: 16 × 64-byte sigs). A hard cap at decode/verify entry would be cheap.

**5. `MultisigAuth` bytes are free w.r.t. plasma.** *[both PARTIAL — consistent with existing model]*
Up to 16×64 bytes of authorization ride outside the hash and outside base-plasma pricing. The existing plasma model doesn't price single-sig signature bytes either, so this is consistent — but combined with #4, unbounded malformed auth is the stronger concern.

**6. `MultisigAuth` malleability (griefing surface, likely benign).** *[Codex PARTIAL — downgraded from Claude's original]*
Since `MultisigAuth` is outside the hash and unsigned, a relayer can mangle it without changing the block hash. Codex verified there is **no** node-wide reject cache keyed by hash (`protocol/handler.go:412` only marks per-peer `knownTxs`), so a subsequently relayed valid copy is not suppressed — the feared reject-then-remember failure is not present in the current code. A targeted test (mangled copy rejected → valid same-hash copy accepted) would lock this in.

**7. Selector ordering invariant is comment-only — verified currently true.** *[Codex DISAGREE on defect, defensive assertion optional]*
`filterExpiredMultisigBlocks` relies on same-address blocks arriving in ascending height order. Codex confirmed both selectors guarantee this today (`pillar/content_selector.go:97`; legacy pool emits per-account chains in order). No present defect; an assertion would only protect future refactors.

### Nits

- `MultisigApi.log` (`rpc/api/embedded/multisig.go:17`) is assigned but never used.
- `ValidPolicy` hardcodes the minimum threshold as a literal `2` (`vm/embedded/definition/multisig.go:150`); a named `MinThreshold` constant beside `MinSigners` would keep protocol bounds in one place.
- `.gitignore` additions (`.pipeline`, `docs/notes`) look like local-dev artifacts; consider a global gitignore instead.
- RPC `GetPolicy` returns `(nil, nil)` for a missing record — a defensible JSON-null contract (and documented), just check consistency with the other embedded APIs.

edgepillar:
Thanks for the multisig work. I reviewed `8f2eddb2b4790e022fe8057aac739c4c99788e8d` and found two issues worth addressing before merge.

**1. [P1] Chain acceptance can depend on a cached account-block patch**

The account verifier describes MA freshness as a node-local admission pre-filter and rejects when `MA + MultisigMaxMaLag < frontierHeight`. However, the current momentum-content verifier checks live authorization without enforcing that MA bound. See [account verification](https://github.com/zenon-network/go-zenon/blob/8f2eddb2b4790e022fe8057aac739c4c99788e8d/verifier/account_block.go#L354) and [momentum-content verification](https://github.com/zenon-network/go-zenon/blob/8f2eddb2b4790e022fe8057aac739c4c99788e8d/verifier/momentum.go#L198).

These checks are not isolated to admission: [`InsertChain`](https://github.com/zenon-network/go-zenon/blob/8f2eddb2b4790e022fe8057aac739c4c99788e8d/protocol/chain_bridge.go#L204) calls `ApplyBlock` for uncached account blocks but skips it when `GetPatch` succeeds. `ApplyBlock` runs the account verifier, including the MA pre-filter.

The cached patch can survive ordinary momentum insertion: [`accountPool.rebuild`](https://github.com/zenon-network/go-zenon/blob/8f2eddb2b4790e022fe8057aac739c4c99788e8d/chain/account_pool.go#L289) preserves still-uncommitted blocks and copies their patches without age revalidation. The producer filter excludes stale blocks from selection; it does not evict their patches.

A concrete differential scenario is:

1. Two nodes have identical committed state and an unchanged multisig policy.
2. Only node A admits an otherwise-valid ordinary multisig transfer while its MA is fresh.
3. Both import the same valid momentums excluding that transfer until frontier `F > MA + MultisigMaxMaLag`.
4. Both receive the same otherwise-valid, correctly produced momentum containing the transfer.

The source paths predict that A bypasses the MA rejection using its cached patch, while B rejects with `ErrABMATooOld`. The [momentum VM](https://github.com/zenon-network/go-zenon/blob/8f2eddb2b4790e022fe8057aac739c4c99788e8d/vm/vm.go#L310) applies the cached patch without repeating account verification.

The default producer filter would omit this transfer; the final momentum requires deliberate construction by an eligible producer. This is not a claim that ordinary default production necessarily triggers the issue.

Could import enforce the same deterministic validity rules regardless of cache state? If MA freshness is intentionally admission-only, keep that filter outside chain replay while preserving structural and authorization checks. A paired cached/uncached `InsertChain` test should cover the lag boundary and beyond, asserting identical acceptance and committed state.

**2. [P2] Nil validation now occurs after dereference**

The [new authorization loop](https://github.com/zenon-network/go-zenon/blob/8f2eddb2b4790e022fe8057aac739c4c99788e8d/verifier/momentum.go#L205) accesses `block.Address` before the [existing nil-block guard](https://github.com/zenon-network/go-zenon/blob/8f2eddb2b4790e022fe8057aac739c4c99788e8d/verifier/momentum.go#L231). A nil element therefore panics instead of returning the intended error.

Please validate elements before authorization and test nil both first and after a valid element. I have not established that the network decoder can produce this input, so I am not claiming a remotely reachable crash.

This was a static source review: I did not run tests or independently verify successful CI.

openmac0815:
**Review — approve in principle, with pre-activation requirements**

Scope-checked the full file list and the gating. The MultisigSpork placeholder is correctly dormant (no on-chain spork can match it until the governance wiring lands), and `GetEmbeddedMethod` gating matches the established DynamicPlasma pattern — the unconditional `applyMultisigDiffs` in `getAllEmbedded` is registration for listing, same as the other contracts, not an activation bypass. So merging this changes no live consensus behavior.

That said, this is 5k+ lines touching verifier signature policy, an embedded contract, and wallet plumbing. Before the spork is ever activated I'd want: (1) a dedicated cryptographic review of the aggregate/policy verification in `verifier/account_block.go` and `wallet/multisig.go`; (2) a devnet soak with the conformance suite. Recommend merging only with the explicit understanding that activation is a separate, gated decision.

PR #73, opened 2026-08-09
libp2p & Dynamic Plasma

digitalSloth:
# libp2p & Dynamic Plasma

This PR promotes the accumulated `dev` work to `testnet` ahead of spinning up a new testnet. It bundles two major features — a networking-layer migration and a new plasma pricing model — plus a follow-up fix for each.

## Highlights

- **libp2p migration**, replacing the custom devp2p/RLPX networking layer with [libp2p](https://libp2p.io/), gated behind a spork so nodes hot-swap transport in-process with no restart required.
- **Dynamic Plasma**, replacing the fixed plasma pricing model with an adaptive, EIP-1559-style mechanism that adjusts fusion and PoW prices each momentum based on network utilization.
- A fix restoring the libp2p spork's registration in `ImplementedSporksMap` after it was dropped in an earlier rebase.
- Multiple correctness fixes closing out review findings on the dynamic plasma work (overflow hardening, canonical base plasma recomputation, duplicate-block handling, big.Int arithmetic for price clamps).

---

### libp2p migration (#59, fixed by #71)

Replaces the custom devp2p/RLPX networking layer with libp2p. The transition is **spork-gated**: every node ships a binary containing both transport stacks, runs on legacy until the libp2p activation spork's `EnforcementHeight` is reached on its local chain, then atomically swaps to libp2p in-process — no restart required by operators.

- New `p2p/libp2p/` backend: host, DHT discovery, Noise security, stream handler, identity conversion, bootstrap parsing.
- New `p2p/switcher/` package: a spork-gated wrapper server that owns both backends, polls a `SporkOracle`, and hot-swaps at `EnforcementHeight`. It implements `p2p.Server`, so `node.go`, `rpc/`, and `protocol/` see a single server type and never reach the active backend directly.
- Legacy devp2p/RLPX code relocated to `p2p/legacy/` (package rename only); `p2p/discover/` and shared message/peer-error types stay at root, used by both backends.
- Node identity is preserved: the existing `~/.znn/network-private-key` (secp256k1 ECDSA) is reused unchanged by both backends. Legacy derives a raw-pubkey `NodeID`; libp2p derives a CIDv1-multihash `peer.ID` (`16Uiu2HAm…`) from the same key. External systems that store or compare old `NodeID` values will need updating after activation.
- Adds peerstore persistence (LevelDB-backed), NAT port mapping support, and DHT discovery hardening, plus lifecycle and identity/bootstrap-parsing unit tests.

**#71** restores `Libp2pSpork` to `ImplementedSporksMap`, which had been dropped in an earlier rebase — without it, `GotAllActiveSporksImplemented` would flag an activated libp2p spork as unimplemented and halt the node.

---

### Dynamic Plasma (#63, hardened by follow-up commits)

Replaces the fixed plasma pricing model with an adaptive one, similar to how Ethereum's EIP-1559 adjusts base fees. Fusion and PoW prices adjust each momentum based on network utilization. Activates behind the `DynamicPlasmaSpork` hard consensus change.

- New `dp/dp.go`: core price-calculation logic — each momentum carries `nextFusionPrice` and `nextWorkPrice`, starting at `MinResourcePrice`. After each momentum, prices adjust by `currentPrice * (usedPlasma - targetPlasma) / targetPlasma / priceChangeDenominator`, capped at 10% change by default to prevent volatility.
- New `pillar/content_selector.go`: ranks blocks for momentum inclusion by `blockValue / BasePlasma`, where `blockValue = FusedPlasma * workPrice + DifficultyToPlasma(Difficulty) * fusionPrice`, filling momentums up to `maxBasePlasmaInMomentum` instead of the old flat 100-block cap. In practice this raises the per-momentum ceiling to roughly 200 simple user sends or 80 contract interactions at default settings, and the limit is configurable via the Plasma contract's `SetVariables` method.
- Core changes span momentum structure, verifier logic, VM plasma calculations, RPC endpoints, and embedded contract variables; the protocol and legacy P2P layers are unmodified by this feature.
- Full design and testing documentation: [`docs/DYNAMIC_PLASMA.md`](../blob/dev/docs/DYNAMIC_PLASMA.md).

**Follow-up hardening** addresses review findings on top of the initial merge:
- Use `big.Int` arithmetic for dynamic-plasma price clamps to avoid overflow at extreme prices.
- Harden the fusion required-plasma calculation against overflow.
- Recompute canonical base plasma in the momentum verifier rather than trusting producer-supplied values.
- Consume prefetched blocks by identifier and address in the momentum content check, and isolate duplicate-content-header handling from chain-order validation, closing edge cases where a dropped contract batch could leave addresses inconsistently marked.
- Remove a producer priority-address ordering bypass and correct plasma constant/comparison logic.
- Read the dynamic plasma spork status from the chain cache.
- Additional unit tests closing coverage gaps in the pricing fixes.

---

## Testing notes

Devnet end-to-end testing was performed for the libp2p swap (legacy → libp2p transition at spork `EnforcementHeight`, verified via node logs, disappearance of the UDP discovery listener, and peer ID format change from hex to base58) and for dynamic plasma (documented in `docs/DYNAMIC_PLASMA.md`). Recommend a fresh devnet/testnet resync plus a full `go test ./...` on the merged branch before cutting the testnet genesis.


0x3639:
## Follow-up review

I found two additional edge cases that should be addressed before merge.

### 1. `MaxPeers` is enforced after the libp2p connection is established

[`[p2p/libp2p/server.go](https://github.com/zenon-network/go-zenon/blob/047959c8e1dc0603a0d7976c252d0ca3c66e01ba/p2p/libp2p/server.go#L472)`](https://github.com/zenon-network/go-zenon/blob/047959c8e1dc0603a0d7976c252d0ca3c66e01ba/p2p/libp2p/server.go#L472) uses libp2p’s default connection manager. The configured `MaxPeers` and `MaxPendingPeers` are checked only after a peer opens `/znn/eth/1`.

A peer can therefore establish a Noise/libp2p connection and never open the Zenon stream. These connections do not appear in `peerMap`, bypass the configured limits, and compete with legitimate peers for the default libp2p connection-manager slots.

At minimum, configure a connection manager from the server limits and protect peers only after they complete the Zenon handshake:

```go
import (
    connmgr "github.com/libp2p/go-libp2p/p2p/net/connmgr"
)

const zenonPeerProtectionTag = "zenon-protocol"

func (srv *Server) connectionLimits() (low, high int) {
    maxPeers := srv.MaxPeers
    if maxPeers <= 0 {
        maxPeers = p2p.DefaultMaxPeers
    }

    maxPending := srv.MaxPendingPeers
    if maxPending <= 0 {
        maxPending = p2p.DefaultMaxPendingPeers
    }

    return maxPeers, maxPeers + maxPending
}
```

In `Start`:

```go
low, high := srv.connectionLimits()
connectionManager, err := connmgr.NewConnManager(
    low,
    high,
    connmgr.WithGracePeriod(frameReadTimeout),
)
if err != nil {
    return fmt.Errorf("create connection manager: %w", err)
}

opts := []libp2p.Option{
    libp2p.Identity(privKey),
    libp2p.ListenAddrs(listenMaddr),
    libp2p.Security(noise.ID, noise.New),
    libp2p.Muxer("/yamux/1.0.0", yamux.DefaultTransport),
    libp2p.Transport(libp2ptcp.NewTCPTransport),
    libp2p.ConnectionManager(connectionManager),
}
```

After successful inbound or outbound adoption:

```go
srv.host.ConnManager().Protect(remotePeer, zenonPeerProtectionTag)
```

When `runPeer` exits:

```go
if pid := p.remotePeer(); pid != "" {
    srv.host.ConnManager().Unprotect(pid, zenonPeerProtectionTag)
}
```

I would also expire connections that have not completed the Zenon handshake within `frameReadTimeout`. The connection manager provides bounded pruning, while the timeout prevents idle non-Zenon connections from remaining indefinitely.

Please add a regression test that connects raw libp2p hosts without opening `/znn/eth/1` and verifies they cannot displace adopted Zenon peers or exceed the configured connection allowance.

### 2. The uncommitted-block limit rejects replacements

[`[chain/account_pool.go](https://github.com/zenon-network/go-zenon/blob/047959c8e1dc0603a0d7976c252d0ca3c66e01ba/chain/account_pool.go#L168)`](https://github.com/zenon-network/go-zenon/blob/047959c8e1dc0603a0d7976c252d0ca3c66e01ba/chain/account_pool.go#L168) checks the prospective block count before determining whether the transaction is an append, duplicate, or rollback replacement.

At 500 uncommitted blocks, a higher-priced replacement is rejected even though replacing an earlier block rolls back descendants and does not increase the pool length. This can strand an account whose pending chain became underpriced under dynamic plasma.

The count check should apply only to the fast-forward path:

```go
// Fast-forward insertion grows the uncommitted account chain, so the
// per-account length limit applies only on this path. Duplicates and
// rollback replacements do not increase the resulting chain length.
if previous == frontierIdentifier {
    if !forceAdd && !types.IsEmbeddedAddress(address) {
        if err := ap.checkUncommittedBlocksCount(address); err != nil {
            return err
        }
    }

    log.Info("fast-forward inserting account-block")
    return ap.getAccountManager(address).Add(transaction)
}

// Already inserted.
trueBlock, err := frontier.ByHeight(identifier.Height)
// Existing duplicate/replacement logic continues here.
```

Suggested regression tests:

```go
func TestAccountPoolAllowsReplacementAtUncommittedLimit(t *testing.T)
func TestAccountPoolDuplicateAtUncommittedLimitIsIdempotent(t *testing.T)
func TestAccountPoolRejectsAppendPastUncommittedLimit(t *testing.T)
```

The replacement test should fill an account to `MaxUncommittedBlocksPerAccount`, then submit a higher-priced block at an earlier height and verify that the rollback/replacement succeeds.

digitalSloth:
@0x3639 Thanks for the review, Iv opened a new PR to address the issues and replied there - https://github.com/zenon-network/go-zenon/pull/74

PR #74, opened 2026-08-13
fix: lilp2p & Dynamic Plasma PR 

digitalSloth:
This PR addresses the comments along with some additional findings from https://github.com/zenon-network/go-zenon/pull/73

### MaxPeers / MaxPendingPeers enforced only after the Zenon handshake

Confirmed. `p2p/libp2p/server.go` built the libp2p host with no `ConnectionManager`, so `MaxPeers`/`MaxPendingPeers` only ever looked at `peerMap`, which is populated *after* the `/znn/eth/1` handshake completes — raw connections that never spoke Zenon were invisible to both limits. Worse, `MaxPendingPeers` turned out to be a single global counter held per *stream* for the full 30s handshake read timeout, so one peer opening a handful of stalled streams could exhaust the global pending budget and deny every other inbound handshake on the node.

Fixed:
- Capped concurrent pending handshakes at one per remote `peer.ID` (`handleStream` in `server.go`), so a single connection can no longer hold more than one pending slot no matter how many streams it opens.
- Cut the handshake read deadline from the steady-state 30s down to 5s (new `handshakeTimeout` constant; `stream_rw.go` gained `ReadMsgLimitedTimeout` to support it).
- `ClosePeer` on handshake failure, so a retry needs a fresh TCP connection rather than reusing the same one indefinitely.
- Added a libp2p `ConnectionManager` sized from `MaxPeers`/`MaxPendingPeers`, and `Protect()`/`Unprotect()` adopted Zenon peers on adoption/disconnect, so libp2p's own trimmer can no longer evict a real peer to make room for an unauthenticated connection sitting inside the default grace period.
- Regression test: `TestPendingHandshakeCapPerPeer` (`p2p/libp2p/server_test.go`) — a raw libp2p host opens 5 stalled streams on one connection; a legitimate peer must still be able to complete its handshake concurrently. Verified it fails on the old code and passes on the fix.

### Uncommitted-block cap rejects valid replacements

Confirmed. `chain/account_pool.go`'s `checkUncommittedBlocksCount` ran unconditionally before branching into fast-forward / duplicate / rollback, so at the 500-block cap it rejected both a higher-priced replacement (which *shortens* the pending chain via rollback) and an idempotent resubmission of an already-inserted block.

Fixed:
- Moved the check inside the fast-forward branch only (`addAccountBlockTransaction` in `account_pool.go`) — replacements and duplicates are no longer gated by it at all, since neither grows the uncommitted chain.
- Dropped the equivalent pre-check in `rpc/api/ledger.go`'s `PublishRawTransaction`: it ran *before* the account-pool insert and would have kept blocking valid replacements/duplicates at the RPC layer even after the pool-level fix, since it couldn't tell fast-forward from replacement either. The pool's own check is synchronous and authoritative for every call path, so this was purely redundant.
- Removed the now-unused `CheckUncommittedBlocksCount` method from the `Chain` interface and its implementation.
- Regression test: `TestAccountPool_ReplacementAndDuplicateAcceptedAtCap` (`chain/account_pool_test.go`) — fills an account to the cap, then verifies both a duplicate resubmission and a same-height higher-priced replacement are accepted. Verified it fails on the old ordering and passes on the fix.

## Additional issues found while reviewing

- **Pool replacement ranking mismatch** (`chain/account_pool.go`) — `higherPriority` still used the legacy `TotalPlasma`/`BasePlasma` ratio for replacement decisions even after dynamic plasma activates, disagreeing with the `dp.HigherPrice`-based rule that actually decides momentum inclusion. Made it spork-aware: once `DynamicPlasmaSpork` is active it defers to `dp.HigherPrice` (via a widened `Stable` interface exposing `GetFrontierMomentumStore`), falling back to the legacy ratio pre-activation or when the store can't answer (e.g. genesis construction).

- **`GetEmbeddedMethod` rebuilt the whole contract map on every call** (`vm/embedded/embedded.go`) — this sits on the per-block plasma path (including the new `CanonicalBasePlasma` check, which calls it once per block during momentum verification), and was reallocating ~10 contracts / ~60 method structs every time. Memoized all 16 spork-combination variants behind a `sync.Once`, so a call is now an array lookup.

- **Content-selector comparator wasn't a strict weak ordering** (`pillar/content_selector.go`) — `higherPriority` mixed a same-address height rule with a cross-address price rule, which admits 3-cycles (a cheap low-height block and an expensive high-height block from the same address can each rank differently against a third address' block priced between them). `sort.SliceStable` over that comparator could separate a block from its own ancestor, leaving a gap the producer's own chain-order check would then reject the whole momentum for. Rewrote `sortBlocksByPriority` to group blocks by address first (each group ordered by height), then order the *groups* by their lowest-height block's price — this makes the comparator total. Added `TestContent_sortBlocksByPriority_NoGapAcrossThirdAddress`, verified it fails on the old comparator and passes on the fix.

- **Plasma-availability error converted into a panic** (`vm/vm.go`) — `enoughPlasma` switched a `return err` into `common.DealWithErr(err)`, turning a clean block rejection (including the data-dependent "got negative available plasma" case) into a hard panic on a path that processes every remote account block. Restored `return err`.

- **`ValidPrice` uint64 overflow** (`dp/dp.go`) — `AccountBlockBasePlasma * fusionPrice` was computed in raw uint64, which overflows and silently wraps the minimum-price floor down to a small value at extreme (governance-settable) prices — exactly defeating the check when it matters most. Now computed in `big.Int` and saturated at `MaxUint64` on overflow (fail closed instead of fail open). Added `TestValidPrice` and `TestValidPrice_OverflowSaturatesInsteadOfWrappingToZero`.

- **Verifier mutating its input, and a dropped size pre-check** (`verifier/momentum.go`) — `content()` wrote the recomputed canonical `BasePlasma` directly into the caller's block (harmless today since nothing downstream reads it afterward, but a landmine for any future caller that shares block pointers). It now uses a local shallow copy for its own price checks instead. Also restored a cheap content-length bound on the dynamic-plasma path *before* building the lookup maps/slices — previously only the legacy path had an upfront size check, so the dynamic-plasma path allocated at attacker-supplied content size before any bound applied. Added `TestRawMomentumVerifier_Content_DoesNotMutateInputBlock` and `TestRawMomentumVerifier_Content_DynamicPlasma_OversizedContentRejected`, both verified against the pre-fix code.

- **Peer database unbounded growth, pure-recency seed selection** (`p2p/libp2p/peerdb.go`) — every handshaked peer was recorded with no cap on total records, and `seeds()` ranked candidates purely by `lastSeen`, letting a source that cycles many identities through the handshake dominate a restarting node's warm-bootstrap dial list. Capped the database at 2000 records (oldest evicted during the existing hourly `expire()` pass), and reworked `seeds()` to group candidates by /24 (IPv4) or /48 (IPv6) subnet and select round-robin across subnets rather than taking a pure recency-ranked prefix. Added `TestPeerDBExpireCapsRecordCount` and `TestPeerDBSeedsDiversifyAcrossSubnets`.

- **Formatting / naming cleanup** — `gofmt -w` on the files this PR touched that weren't clean (`common/types/spork.go`, `p2p/legacy/nat/nat.go`, `p2p/legacy/peer.go`, `p2p/legacy/server.go`, `vm/embedded/tests/dp_test.go`, `vm/embedded/tests/z_bridge_test.go`); renamed the bug-numbered switcher tests (`TestStopPreActivation_Bug1` → `TestStopDoesNotDeadlockBeforeActivation`, etc.) to describe behaviour instead of a past review round; moved `getAllEmbedded()` out of production code into `embedded_test.go`, since it exists solely for test introspection.

## Verification

Every fix was checked with `go build`/`go vet` plus the relevant package's existing test suite, and I added regression tests for the ones above where it was practical to write one — for the higher-severity fixes (P2P handshake cap, account-pool cap ordering, content-selector cycle, `ValidPrice` overflow, verifier mutation, peer-DB cap/diversity) I also confirmed each new test actually discriminates: it fails against the pre-fix code and passes against the fix, not just "returns no error" either way. Full repo test suite is clean (excluding the pre-existing, unrelated `p2p/legacy/nat` UPnP test hang on macOS that's already flagged in this thread).

One more item from the review I'm handling directly rather than through this PR: `GovernanceAddress`. I'll follow up separately once I've settled on the right address for testnet.

### DIFF
--- a/chain/account_pool.go
+++ b/chain/account_pool.go
@@ -153,7 +153,7 @@ func higherPricedBlock(plasma dp.DynamicPlasma, a, b *nom.AccountBlock) error {
 // replacement that content selection would then rank below the block it
 // just evicted. An exact price tie is resolved here by smallest hash, the
 // fork-resolution rule documented on the AccountPool interface
-// (chain/interface.go:53-55), deliberately independent of the content
+// (chain/interface.go), deliberately independent of the content
 // selector's larger-hash tie-break: the two answer different questions —
 // which block survives a fork vs. which block is emitted first. Pre-
 // activation (or if the frontier momentum store can't answer, e.g. in
--- a/chain/account_pool.go
+++ b/chain/account_pool.go
@@ -130,23 +130,46 @@ func (ap *accountPool) canRollback(block *nom.AccountBlock) error {
 	return nil
 }
 
+// higherPricedBlock reports whether a should replace b under dynamic-plasma
+// pricing. An exact price tie is resolved by smallest hash so that nodes
+// seeing the two blocks in different orders converge on the same one, the
+// rule documented on the AccountPool interface.
+func higherPricedBlock(plasma dp.DynamicPlasma, a, b *nom.AccountBlock) error {
+	err := plasma.HigherPrice(a, b)
+	if err == dp.ErrBlockPriceSame {
+		if bytes.Compare(a.Hash.Bytes()[:], b.Hash.Bytes()[:]) > -1 {
+			return ErrHashTieBreak
+		}
+		return nil
+	}
+	return err
+}
+
 // higherPriority reports whether a should replace b as the pool's block
 // at their shared height. Once dynamic plasma is active, momentum
-// content selection ranks blocks by dp.DynamicPlasma.HigherPrice (see
-// pillar/content_selector.go), so replacement must use the same rule —
-// otherwise the pool can accept a replacement that content selection
-// would then rank below the block it just evicted, or reject one it
-// would have preferred. Pre-activation (or if the frontier momentum
-// store can't answer, e.g. in genesis construction), momentum content
-// selection still uses the legacy TotalPlasma/BasePlasma ratio, so that
-// remains the fallback.
+// content selection ranks blocks by the same price rule,
+// dp.DynamicPlasma.HigherPrice (see pillar/content_selector.go), so
+// replacement must use it too — otherwise the pool can accept a
+// replacement that content selection would then rank below the block it
+// just evicted. An exact price tie is resolved here by smallest hash, the
+// fork-resolution rule documented on the AccountPool interface
+// (chain/interface.go:53-55), deliberately independent of the content
+// selector's larger-hash tie-break: the two answer different questions —
+// which block survives a fork vs. which block is emitted first. Pre-
+// activation (or if the frontier momentum store can't answer, e.g. in
+// genesis construction), momentum content selection still uses the legacy
+// TotalPlasma/BasePlasma ratio, so that remains the fallback.
 func (ap *accountPool) higherPriority(a, b *nom.AccountBlock) error {
 	if store := ap.stable.GetFrontierMomentumStore(); store != nil {
-		if active, err := store.IsSporkActive(types.DynamicPlasmaSpork); err == nil && active {
-			if previous, err := store.GetFrontierMomentum(); err == nil {
-				if config, err := store.GetPlasmaVariables(); err == nil {
-					return dp.NewDynamicPlasma(previous, config).HigherPrice(a, b)
-				}
+		if active, err := store.IsSporkActive(types.DynamicPlasmaSpork); err != nil {
+			ap.log.Debug("using legacy plasma-ratio comparator", "reason", err)
+		} else if active {
+			if previous, err := store.GetFrontierMomentum(); err != nil {
+				ap.log.Debug("using legacy plasma-ratio comparator", "reason", err)
+			} else if config, err := store.GetPlasmaVariables(); err != nil {
+				ap.log.Debug("using legacy plasma-ratio comparator", "reason", err)
+			} else {
+				return higherPricedBlock(dp.NewDynamicPlasma(previous, config), a, b)
 			}
 		}
 	}
--- a/chain/account_pool.go
+++ b/chain/account_pool.go
@@ -14,6 +14,7 @@ import (
 	"github.com/zenon-network/go-zenon/common"
 	"github.com/zenon-network/go-zenon/common/db"
 	"github.com/zenon-network/go-zenon/common/types"
+	"github.com/zenon-network/go-zenon/dp"
 )
 
 var (
@@ -33,6 +34,7 @@ var (
 
 type Stable interface {
 	GetStableAccountDB(address types.Address) db.DB
+	GetFrontierMomentumStore() store.Momentum
 }
 
 type accountManager struct {
@@ -128,7 +130,27 @@ func (ap *accountPool) canRollback(block *nom.AccountBlock) error {
 	return nil
 }
 
-func higherPriority(a, b *nom.AccountBlock) error {
+// higherPriority reports whether a should replace b as the pool's block
+// at their shared height. Once dynamic plasma is active, momentum
+// content selection ranks blocks by dp.DynamicPlasma.HigherPrice (see
+// pillar/content_selector.go), so replacement must use the same rule —
+// otherwise the pool can accept a replacement that content selection
+// would then rank below the block it just evicted, or reject one it
+// would have preferred. Pre-activation (or if the frontier momentum
+// store can't answer, e.g. in genesis construction), momentum content
+// selection still uses the legacy TotalPlasma/BasePlasma ratio, so that
+// remains the fallback.
+func (ap *accountPool) higherPriority(a, b *nom.AccountBlock) error {
+	if store := ap.stable.GetFrontierMomentumStore(); store != nil {
+		if active, err := store.IsSporkActive(types.DynamicPlasmaSpork); err == nil && active {
+			if previous, err := store.GetFrontierMomentum(); err == nil {
+				if config, err := store.GetPlasmaVariables(); err == nil {
+					return dp.NewDynamicPlasma(previous, config).HigherPrice(a, b)
+				}
+			}
+		}
+	}
+
 	if a.TotalPlasma*b.BasePlasma < b.TotalPlasma*a.BasePlasma {
 		return ErrPlasmaRatioIsWorse
 	} else if a.TotalPlasma*b.BasePlasma == b.TotalPlasma*a.BasePlasma && bytes.Compare(a.Hash.Bytes()[:], b.Hash.Bytes()[:]) > -1 {
@@ -165,15 +187,17 @@ func (ap *accountPool) addAccountBlockTransaction(transaction *nom.AccountBlockT
 	frontier := ap.getFrontierAccountStore(address)
 	frontierIdentifier := frontier.Identifier()
 
-	// check uncommitted plasma amount
-	if !forceAdd && !types.IsEmbeddedAddress(address) {
-		if err := ap.checkUncommittedBlocksCount(address); err != nil {
-			return err
-		}
-	}
-
 	// fast-forward insert on top of chain
 	if previous == frontierIdentifier {
+		// check uncommitted plasma amount. Only applies to fast-forward
+		// inserts: a rollback/replacement doesn't lengthen the pending
+		// chain, and an already-inserted duplicate is idempotent —
+		// neither should be rejected just because the account is at cap.
+		if !forceAdd && !types.IsEmbeddedAddress(address) {
+			if err := ap.checkUncommittedBlocksCount(address); err != nil {
+				return err
+			}
+		}
 		log.Info("fast-forward inserting account-block")
 		return ap.getAccountManager(address).Add(transaction)
 	}
@@ -192,7 +216,7 @@ func (ap *accountPool) addAccountBlockTransaction(transaction *nom.AccountBlockT
 	if err := ap.canRollback(block); err != nil {
 		return err
 	}
-	if err := higherPriority(block, trueBlock); !forceAdd && err != nil {
+	if err := ap.higherPriority(block, trueBlock); !forceAdd && err != nil {
 		log.Info("failed to insert account-block-transaction", "reason", err, "frontier-identifier", frontierIdentifier)
 		return err
 	}
@@ -376,12 +400,6 @@ func (ap *accountPool) getUncommittedAccountBlocksByAddress(address types.Addres
 	return blocks
 }
 
-func (ap *accountPool) CheckUncommittedBlocksCount(address types.Address) error {
-	ap.changes.Lock()
-	defer ap.changes.Unlock()
-
-	return ap.checkUncommittedBlocksCount(address)
-}
 func (ap *accountPool) checkUncommittedBlocksCount(address types.Address) error {
 	frontier, err := ap.getFrontierAccountStore(address).Frontier()
 	if err != nil {
--- a/chain/account_pool.go
+++ b/chain/account_pool.go
@@ -80,6 +80,22 @@ type accountPool struct {
 	stable   Stable
 	managers map[types.Address]*accountManager
 	changes  sync.Mutex
+
+	// plasma is the dynamic-plasma pricing context derived from the
+	// committed frontier momentum, or nil while the dynamic-plasma spork
+	// is inactive or the momentum store cannot answer. It only changes
+	// when the frontier momentum does, so it is refreshed once per
+	// committed (or rolled-back) momentum rather than recomputed on every
+	// contested insert.
+	//
+	// plasmaMu guards the field alone: it is never held across a store
+	// read or across ap.changes, so it is a leaf lock. Every production
+	// writer (momentum insert/rollback) and reader (higherPriority) runs
+	// under chain.insert already, so plasmaMu is what keeps that invariant
+	// enforced rather than assumed — including for direct unit-test use of
+	// the pool, and for -race.
+	plasma   dp.DynamicPlasma
+	plasmaMu sync.Mutex
 }
 
 func (ap *accountPool) getAccountManager(address types.Address) *accountManager {
@@ -156,22 +172,17 @@ func higherPricedBlock(plasma dp.DynamicPlasma, a, b *nom.AccountBlock) error {
 // (chain/interface.go), deliberately independent of the content
 // selector's larger-hash tie-break: the two answer different questions —
 // which block survives a fork vs. which block is emitted first. Pre-
-// activation (or if the frontier momentum store can't answer, e.g. in
-// genesis construction), momentum content selection still uses the legacy
+// activation, the comparator uses the pricing context cached from the
+// committed frontier momentum; while that context is empty (dynamic
+// plasma inactive, or the momentum store could not answer — genesis
+// construction, for one) momentum content selection still uses the legacy
 // TotalPlasma/BasePlasma ratio, so that remains the fallback.
 func (ap *accountPool) higherPriority(a, b *nom.AccountBlock) error {
-	if store := ap.stable.GetFrontierMomentumStore(); store != nil {
-		if active, err := store.IsSporkActive(types.DynamicPlasmaSpork); err != nil {
-			ap.log.Debug("using legacy plasma-ratio comparator", "reason", err)
-		} else if active {
-			if previous, err := store.GetFrontierMomentum(); err != nil {
-				ap.log.Debug("using legacy plasma-ratio comparator", "reason", err)
-			} else if config, err := store.GetPlasmaVariables(); err != nil {
-				ap.log.Debug("using legacy plasma-ratio comparator", "reason", err)
-			} else {
-				return higherPricedBlock(dp.NewDynamicPlasma(previous, config), a, b)
-			}
-		}
+	ap.plasmaMu.Lock()
+	plasma := ap.plasma
+	ap.plasmaMu.Unlock()
+	if plasma != nil {
+		return higherPricedBlock(plasma, a, b)
 	}
 
 	if a.TotalPlasma*b.BasePlasma < b.TotalPlasma*a.BasePlasma {
@@ -305,7 +316,52 @@ func (ap *accountPool) getFrontierAccountStore(address types.Address) store.Acco
 	return account.NewAccountStore(address, ap.getAccountManager(address).db.Frontier())
 }
 
+// refreshDynamicPlasma recomputes the pricing context from the committed
+// frontier momentum. Called on every momentum insert and rollback, and
+// once at chain start-up.
+func (ap *accountPool) refreshDynamicPlasma() {
+	plasma := ap.computeDynamicPlasma()
+	ap.plasmaMu.Lock()
+	ap.plasma = plasma
+	ap.plasmaMu.Unlock()
+}
+
+// computeDynamicPlasma returns the pricing context for the committed
+// frontier momentum, or nil when dynamic plasma is inactive or the
+// momentum store cannot answer. A store that cannot answer is logged at
+// Warn: it means pool replacement falls back to the legacy plasma-ratio
+// comparator while momentum content selection keeps ranking by price, so
+// the pool can keep a block content selection would rank below the one it
+// evicted.
+func (ap *accountPool) computeDynamicPlasma() dp.DynamicPlasma {
+	store := ap.stable.GetFrontierMomentumStore()
+	if store == nil {
+		return nil
+	}
+	active, err := store.IsSporkActive(types.DynamicPlasmaSpork)
+	if err != nil {
+		ap.log.Warn("using legacy plasma-ratio comparator", "reason", err)
+		return nil
+	}
+	if !active {
+		return nil
+	}
+	previous, err := store.GetFrontierMomentum()
+	if err != nil {
+		ap.log.Warn("using legacy plasma-ratio comparator", "reason", err)
+		return nil
+	}
+	config, err := store.GetPlasmaVariables()
+	if err != nil {
+		ap.log.Warn("using legacy plasma-ratio comparator", "reason", err)
+		return nil
+	}
+	return dp.NewDynamicPlasma(previous, config)
+}
+
 func (ap *accountPool) InsertMomentum(detailed *nom.DetailedMomentum) {
+	ap.refreshDynamicPlasma()
+
 	ap.changes.Lock()
 	defer ap.changes.Unlock()
 
@@ -314,6 +370,8 @@ func (ap *accountPool) InsertMomentum(detailed *nom.DetailedMomentum) {
 	}
 }
 func (ap *accountPool) DeleteMomentum(*nom.DetailedMomentum) {
+	ap.refreshDynamicPlasma()
+
 	ap.changes.Lock()
 	defer ap.changes.Unlock()
 
--- a/chain/account_pool.go
+++ b/chain/account_pool.go
@@ -171,12 +171,12 @@ func higherPricedBlock(plasma dp.DynamicPlasma, a, b *nom.AccountBlock) error {
 // fork-resolution rule documented on the AccountPool interface
 // (chain/interface.go), deliberately independent of the content
 // selector's larger-hash tie-break: the two answer different questions —
-// which block survives a fork vs. which block is emitted first. Pre-
-// activation, the comparator uses the pricing context cached from the
-// committed frontier momentum; while that context is empty (dynamic
-// plasma inactive, or the momentum store could not answer — genesis
-// construction, for one) momentum content selection still uses the legacy
-// TotalPlasma/BasePlasma ratio, so that remains the fallback.
+// which block survives a fork vs. which block is emitted first. The
+// comparator uses the pricing context cached from the committed frontier
+// momentum; while that context is empty (dynamic plasma inactive, or the
+// momentum store could not answer — genesis construction, for one) momentum
+// content selection still uses the legacy TotalPlasma/BasePlasma ratio, so
+// that remains the fallback.
 func (ap *accountPool) higherPriority(a, b *nom.AccountBlock) error {
 	ap.plasmaMu.Lock()
 	plasma := ap.plasma
--- a/chain/account_pool_test.go
+++ b/chain/account_pool_test.go
@@ -8,6 +8,8 @@ import (
 	"github.com/zenon-network/go-zenon/common"
 	"github.com/zenon-network/go-zenon/common/db"
 	"github.com/zenon-network/go-zenon/common/types"
+	"github.com/zenon-network/go-zenon/dp"
+	"github.com/zenon-network/go-zenon/vm/embedded/definition"
 )
 
 type fakeStable struct{}
@@ -201,4 +203,24 @@ func testHashHeight(height uint64) types.HashHeight {
 		Hash:   types.NewHash(common.Uint64ToBytes(height)),
 		Height: height,
 	}
+}
+
+func TestHigherPricedBlock_TieBreaksOnSmallestHash(t *testing.T) {
+	plasma := dp.NewDynamicPlasma(&nom.Momentum{Version: 2, NextFusionPrice: 1000, NextWorkPrice: 1000}, &definition.PlasmaVariables{})
+
+	smallerHash := &nom.AccountBlock{BasePlasma: 21000, FusedPlasma: 21000}
+	smallerHash.Hash = types.Hash{0}
+	largerHash := &nom.AccountBlock{BasePlasma: 21000, FusedPlasma: 21000}
+	largerHash.Hash = types.Hash{1}
+
+	// Smaller hash wins the tie: a replaces b.
+	common.FailIfErr(t, higherPricedBlock(plasma, smallerHash, largerHash))
+
+	// Larger hash loses the tie: a does not replace b.
+	common.ExpectError(t, higherPricedBlock(plasma, largerHash, smallerHash), ErrHashTieBreak)
+
+	// An unambiguously cheaper block passes the price result through
+	// unchanged; the tie-break never fires.
+	cheaper := &nom.AccountBlock{BasePlasma: 21000, FusedPlasma: 100}
+	common.ExpectError(t, higherPricedBlock(plasma, cheaper, largerHash), dp.ErrBlockPriceWorse)
 }
--- a/chain/account_pool_test.go
+++ b/chain/account_pool_test.go
@@ -4,6 +4,7 @@ import (
 	"testing"
 
 	"github.com/zenon-network/go-zenon/chain/nom"
+	"github.com/zenon-network/go-zenon/chain/store"
 	"github.com/zenon-network/go-zenon/common"
 	"github.com/zenon-network/go-zenon/common/db"
 	"github.com/zenon-network/go-zenon/common/types"
@@ -13,6 +14,13 @@ type fakeStable struct{}
 
 func (fakeStable) GetStableAccountDB(types.Address) db.DB {
 	return db.NewMemDB()
+}
+
+// GetFrontierMomentumStore returns nil so higherPriority falls back to
+// the legacy ratio comparator; these tests don't exercise dynamic
+// plasma activation.
+func (fakeStable) GetFrontierMomentumStore() store.Momentum {
+	return nil
 }
 
 // A never-committed account has no stable frontier, so its uncommitted block
@@ -40,6 +48,73 @@ func TestAccountPool_checkUncommittedBlocksCount_FreshAccount(t *testing.T) {
 	}
 
 	common.ExpectTrue(t, ap.checkUncommittedBlocksCount(address) != nil)
+}
+
+// The uncommitted-block cap must only gate inserts that lengthen the
+// pending chain (fast-forward). A same-height replacement (rollback) or
+// a resubmission of an already-inserted block must still be accepted
+// while the account sits at the cap, since neither grows the
+// uncommitted chain.
+func TestAccountPool_ReplacementAndDuplicateAcceptedAtCap(t *testing.T) {
+	// First byte 0 (types.UserAddrByte) so IsEmbeddedAddress is false and
+	// addAccountBlockTransaction's uncommitted-block-count guard actually
+	// applies — unlike the sibling FreshAccount test above, this test
+	// exercises that guard through addAccountBlockTransaction itself, not
+	// just checkUncommittedBlocksCount directly.
+	address := types.Address{0, 1}
+	ap := newAccountPool(fakeStable{})
+	manager := ap.getAccountManager(address)
+
+	previousHash := types.ZeroHash
+	var block500 *nom.AccountBlock
+	for height := uint64(1); height <= MaxUncommittedBlocksPerAccount; height++ {
+		block := &nom.AccountBlock{
+			Address:      address,
+			BlockType:    nom.BlockTypeUserSend,
+			Height:       height,
+			PreviousHash: previousHash,
+			TotalPlasma:  100,
+			BasePlasma:   100,
+		}
+		block.Hash = block.ComputeHash()
+		common.FailIfErr(t, manager.Add(&nom.AccountBlockTransaction{
+			Block:   block,
+			Changes: db.NewPatch(),
+		}))
+		previousHash = block.Hash
+		block500 = block
+	}
+
+	// Sanity: the account is genuinely at the cap for fast-forward inserts.
+	common.ExpectTrue(t, ap.checkUncommittedBlocksCount(address) != nil)
+
+	// Resubmitting the already-inserted frontier block must stay
+	// idempotent rather than error out because the account is at cap.
+	duplicate := &nom.AccountBlockTransaction{Block: block500.Copy(), Changes: db.NewPatch()}
+	common.FailIfErr(t, ap.addAccountBlockTransaction(duplicate, false))
+
+	// A same-height, higher-priced replacement (distinct hash via Data,
+	// same previous/height as block500) must be accepted through the
+	// rollback path: it doesn't grow the uncommitted chain, so the cap
+	// must not block it.
+	replacement := &nom.AccountBlock{
+		Address:      address,
+		BlockType:    nom.BlockTypeUserSend,
+		Height:       MaxUncommittedBlocksPerAccount,
+		PreviousHash: block500.PreviousHash,
+		Data:         []byte{1},
+		TotalPlasma:  200, // higher TotalPlasma/BasePlasma ratio than block500 wins higherPriority
+		BasePlasma:   100,
+	}
+	replacement.Hash = replacement.ComputeHash()
+	common.FailIfErr(t, ap.addAccountBlockTransaction(&nom.AccountBlockTransaction{
+		Block:   replacement,
+		Changes: db.NewPatch(),
+	}, false))
+
+	frontier, err := ap.getFrontierAccountStore(address).Frontier()
+	common.FailIfErr(t, err)
+	common.ExpectTrue(t, frontier != nil && frontier.Identifier() == replacement.Identifier())
 }
 
 func TestAccountPool_filterBlocksToCommit(t *testing.T) {
--- a/chain/account_pool_test.go
+++ b/chain/account_pool_test.go
@@ -18,9 +18,9 @@ func (fakeStable) GetStableAccountDB(types.Address) db.DB {
 	return db.NewMemDB()
 }
 
-// GetFrontierMomentumStore returns nil so higherPriority falls back to
-// the legacy ratio comparator; these tests don't exercise dynamic
-// plasma activation.
+// GetFrontierMomentumStore returns nil so the pool's pricing context stays
+// empty and higherPriority falls back to the legacy ratio comparator;
+// these tests don't exercise dynamic plasma activation.
 func (fakeStable) GetFrontierMomentumStore() store.Momentum {
 	return nil
 }
@@ -223,4 +223,32 @@ func TestHigherPricedBlock_TieBreaksOnSmallestHash(t *testing.T) {
 	// unchanged; the tie-break never fires.
 	cheaper := &nom.AccountBlock{BasePlasma: 21000, FusedPlasma: 100}
 	common.ExpectError(t, higherPricedBlock(plasma, cheaper, largerHash), dp.ErrBlockPriceWorse)
+}
+
+// TestAccountPool_higherPriorityUsesCachedDynamicPlasma verifies that
+// higherPriority is driven by the pool's cached pricing context, not by a
+// fresh store read: fakeStable's store is nil, yet the price comparator
+// still runs while the cache is set, and stops running once it is cleared.
+func TestAccountPool_higherPriorityUsesCachedDynamicPlasma(t *testing.T) {
+	ap := newAccountPool(fakeStable{}) // its store is nil: no store answer available
+	ap.plasma = dp.NewDynamicPlasma(
+		&nom.Momentum{Version: 2, NextFusionPrice: 1000, NextWorkPrice: 1000},
+		&definition.PlasmaVariables{},
+	)
+
+	cheap := &nom.AccountBlock{BasePlasma: 21000, FusedPlasma: 100}
+	cheap.Hash = types.Hash{1}
+	rich := &nom.AccountBlock{BasePlasma: 21000, FusedPlasma: 21000}
+	rich.Hash = types.Hash{0}
+
+	// With the cache set, the price comparator ran even though fakeStable
+	// has no store.
+	common.ExpectError(t, ap.higherPriority(cheap, rich), dp.ErrBlockPriceWorse)
+
+	// With the cache cleared, both blocks have TotalPlasma == 0, so the
+	// legacy branch's ratio comparison is an equality and resolves on the
+	// hash tie-break, which cheap.Hash{1} > rich.Hash{0} makes
+	// deterministic.
+	ap.plasma = nil
+	common.ExpectError(t, ap.higherPriority(cheap, rich), ErrHashTieBreak)
 }
--- a/chain/account_pool_test.go
+++ b/chain/account_pool_test.go
@@ -252,3 +252,20 @@ func TestAccountPool_higherPriorityUsesCachedDynamicPlasma(t *testing.T) {
 	ap.plasma = nil
 	common.ExpectError(t, ap.higherPriority(cheap, rich), ErrHashTieBreak)
 }
+
+// TestAccountPool_MomentumEventsRefreshDynamicPlasma verifies that
+// InsertMomentum and DeleteMomentum both refresh the pool's cached pricing
+// context by calling refreshDynamicPlasma, not just that higherPriority
+// consults the cache once it is populated.
+func TestAccountPool_MomentumEventsRefreshDynamicPlasma(t *testing.T) {
+	ap := newAccountPool(fakeStable{}) // nil store: any refresh must clear the cache
+	stale := dp.NewDynamicPlasma(&nom.Momentum{Version: 2, NextFusionPrice: 1000, NextWorkPrice: 1000}, &definition.PlasmaVariables{})
+
+	ap.plasma = stale
+	ap.InsertMomentum(&nom.DetailedMomentum{Momentum: &nom.Momentum{}})
+	common.Expect(t, ap.plasma == nil, true)
+
+	ap.plasma = stale
+	ap.DeleteMomentum(nil)
+	common.Expect(t, ap.plasma == nil, true)
+}
--- a/chain/chain.go
+++ b/chain/chain.go
@@ -59,6 +59,11 @@ func (c *chain) Init() error {
 	}
 	types.SporkAddress = c.genesis.GetSporkAddress()
 	c.Register(c.accountPool)
+	// Prime the pool's pricing context from the committed frontier so the
+	// first contested insert after start-up is priced the same way as one
+	// arriving after the next momentum, rather than falling back to the
+	// legacy comparator until then.
+	c.accountPool.refreshDynamicPlasma()
 
 	frontierStore := c.GetFrontierMomentumStore()
 	frontier, err := frontierStore.GetFrontierMomentum()
--- a/chain/genesis/account_block.go
+++ b/chain/genesis/account_block.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/zenon-network/go-zenon/chain"
 	"github.com/zenon-network/go-zenon/chain/nom"
+	"github.com/zenon-network/go-zenon/chain/store"
 	"github.com/zenon-network/go-zenon/common"
 	"github.com/zenon-network/go-zenon/common/db"
 	"github.com/zenon-network/go-zenon/common/types"
@@ -18,6 +19,14 @@ type mockStable struct {
 
 func (m *mockStable) GetStableAccountDB(address types.Address) db.DB {
 	return db.NewMemDB()
+}
+
+// GetFrontierMomentumStore returns nil: genesis constructs the very
+// first account blocks before any momentum exists, so there is no spork
+// state to consult. accountPool.higherPriority falls back to the legacy
+// ratio comparator whenever this is nil.
+func (m *mockStable) GetFrontierMomentumStore() store.Momentum {
+	return nil
 }
 
 func newGenesisAccountBlocks(cfg *GenesisConfig) chain.AccountPool {
--- a/chain/interface.go
+++ b/chain/interface.go
@@ -63,8 +63,6 @@ type AccountPool interface {
 	GetNewMomentumContent() []*nom.AccountBlock
 	GetAllUncommittedAccountBlocks() []*nom.AccountBlock
 	GetUncommittedAccountBlocksByAddress(address types.Address) []*nom.AccountBlock
-
-	CheckUncommittedBlocksCount(address types.Address) error
 }
 
 type ChainCache interface {
--- a/chain/tests/account_pool_test.go
+++ b/chain/tests/account_pool_test.go
@@ -5,6 +5,7 @@ import (
 	"math/big"
 	"testing"
 
+	"github.com/zenon-network/go-zenon/chain"
 	g "github.com/zenon-network/go-zenon/chain/genesis/mock"
 	"github.com/zenon-network/go-zenon/chain/nom"
 	"github.com/zenon-network/go-zenon/common"
@@ -114,7 +115,7 @@ func TestAccountPool_Priority(t *testing.T) {
 	common.Expect(t, len(uncommitted), 1)
 	common.ExpectString(t, uncommitted[0].Hash.String(), highPriorityBlock.Hash.String())
 
-	z.InsertSendBlock(lowPriorityBlock, nil, mock.SkipVmChanges)
+	z.InsertSendBlockRejected(lowPriorityBlock, chain.ErrPlasmaRatioIsWorse)
 
 	uncommitted = z.Chain().GetAllUncommittedAccountBlocks()
 	common.Expect(t, len(uncommitted), 1)
--- a/common/types/spork.go
+++ b/common/types/spork.go
@@ -25,15 +25,15 @@ var (
 	// for the existing pattern).
 
 	// Libp2pSpork gates the activation of the libp2p networking stack.
-    // Until this spork's EnforcementHeight is reached on a node's local
-    // chain, the legacy (devp2p/RLPX) backend is in use; at activation,
-    // every node atomically switches to the libp2p backend.
+	// Until this spork's EnforcementHeight is reached on a node's local
+	// chain, the legacy (devp2p/RLPX) backend is in use; at activation,
+	// every node atomically switches to the libp2p backend.
 	Libp2pSpork = NewImplementedSpork("0000000000000000000000000000000000000000000000000000000000000001")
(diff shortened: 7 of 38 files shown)


0x3639 (commented):
# PR #74 review — fix: libp2p & Dynamic Plasma

These comments were a collaboration between Fable and Codex 5.6.

**Verdict: Request changes.** This PR is a clear net improvement — the per-peer pending-handshake cap, the connection manager with `Protect()` on adoption, the account-pool cap reordering, the embedded-contract memoization, and the `enoughPlasma` panic fix are all correct, and the regression tests that discriminate pre-fix vs post-fix are appreciated. But the content-selector rewrite trades the old ordering bug for a fee-priority bypass, the `ValidPrice` overflow fix is incomplete in one direction, removing the RPC-level cap check silently drops the only user-visible error for a full account pool, and the new handshake-failure `ClosePeer` rests on an incorrect invariant.

---

## 1. [P1] Group-contiguous content selection lets one account bypass per-block price priority

`pillar/content_selector.go` — `sortBlocksByPriority`

The rewrite ranks an entire address group by the price of its *lowest-height* block only, then emits the whole group contiguously. `filterBlocksToCommit` consumes the momentum plasma budget in that order, so every descendant inherits its ancestor's queue position regardless of its own price — the per-block price competition the old comparator provided is not re-established anywhere.

Concretely: submit `h1` priced high plus `h2..h200` at the `ValidPrice` floor. At default config the numbers line up exactly — `DefaultMaxBasePlasmaInMomentum = 4,200,000 = 200 × 21,000`, and the dynamic-plasma path has no per-address or block-count cap (the 100-block `MaxAccountBlocksInMomentum` cap only applies to the legacy path). One premium fee escorts 199 floor-priced blocks into the momentum and crowds out every other user's higher-priced blocks. Under the old comparator, each floor block competed individually on price and the tail was truncated instead. This is producer-local selection policy, not a consensus rule, but once pillars run this code it's the network's fee market under congestion.

The motivation for the rewrite is sound — the old per-block comparator genuinely wasn't a strict weak ordering and could split an address chain. But the fix over-corrects. The standard shape that satisfies both constraints (chain order within an address, price order across addresses) is a priority structure over each address's *currently eligible head*: pop the best head, then its successor becomes that address's new head and competes on its own price. With prices `A1=30000, A2=21000, B1=25000` and budget for two, that selects `[A1, B1]` where the current code selects `[A1, A2]`. It also composes cleanly with the existing budget/skip logic. A cheaper patch — keying each group on its *minimum*-priced block instead of its lowest-height one — closes the escort exploit but still under-serves mixed-price chains.

## 2. [P1] `ValidPrice` saturation still admits underpriced PoW blocks when the floor overflows

`dp/dp.go` — `ValidPrice`

The big.Int fix removed the wrap-to-zero failure, but the saturated proxy is still lossy in one direction. When `AccountBlockBasePlasma * fusionPrice / PriceScaleFactor` overflows, the floor is clamped into a synthetic block with `FusedPlasma = MaxUint64`, whose score in `HigherPrice` is `MaxUint64 · workPrice`. A PoW-carrying block's score is weighted by `fusionPrice` — the very quantity that is huge — so it can beat the truncated floor while being far below the true one.

Verified numerically: with `fusionPrice = 1<<61`, `workPrice = 1000`, the true floor is `21000·2^61/1000 ≈ 4.84e19` fused-plasma-equivalent (score ≈ `4.84e22`), but the saturated proxy scores only `≈ 1.84e22`. A block carrying just 10,000 PoW plasma scores `10^4·2^61 ≈ 2.31e22` — under half the real requirement — and `ValidPrice` returns true. The code comment claiming saturation "fails closed" is therefore incorrect for PoW-weighted blocks and should be corrected along with the code.

Severity context: `nextResourcePrice` has no upper bound, so a fusion price past ~8.8e17 is reachable in principle under sustained congestion, though it's ~15 orders of magnitude above `MinResourcePrice`. Either way the robust fix is simple — don't materialize the floor as a synthetic `AccountBlock` at all. Keep everything in `big.Int` and compare directly:

```
blockScore = FusedPlasma·workPrice + DifficultyToPlasma(Difficulty)·fusionPrice        (big.Int)
floorScore = (AccountBlockBasePlasma·fusionPrice / PriceScaleFactor) · workPrice       (big.Int)
valid      ⇔ blockScore · AccountBlockBasePlasma ≥ floorScore · block.BasePlasma
```

While touching this: the floor value depends only on constants and the immutable `dp.fusionPrice`, so it can be computed once in `NewDynamicPlasma` rather than per block — `ValidPrice` runs for every user block in both momentum verification and content selection.

## 3. [P2] Removing the RPC-level cap check makes `PublishRawTransaction` report success for silently dropped blocks

`rpc/api/ledger.go` — `PublishRawTransaction` / `protocol/broadcaster.go` — `CreateAccountBlock`

The PR body says the pool's own check is "synchronous and authoritative for every call path" — true for the chain, but the error never reaches the RPC caller. `CreateAccountBlock` returns void and only logs insert failures:

```go
err := b.chain.AddAccountBlockTransaction(insert, accountBlockTransaction)
insert.Unlock()
if err != nil {
    b.log.Error("failed to insert own account-block", "reason", err)
    return
}
```

Before this PR, the `CheckUncommittedBlocksCount` pre-check was the only path by which a wallet learned its transaction was rejected for a full pending chain. After it, `ledger.publishRawTransaction` returns success while the block is neither inserted nor broadcast. Worse, the wallet then signs its next block on top of the never-inserted one — which is also silently dropped (`previous != frontier`) — so the account chain stalls with no error ever surfaced, and tooling that retries only on error never resubmits.

The removed pre-check was indeed too blunt (it couldn't distinguish fast-forward from replacement/duplicate — the very bug this PR fixes at the pool layer). But the right move is to propagate, not delete: either have `CreateAccountBlock` return the insert error up through the broadcaster interface, or reinstate a pre-check that only applies when the submitted block is actually a fast-forward (`block.Previous == frontier identifier`), mirroring the pool's new logic.

## 4. [P2] Pre-adoption `ClosePeer` can tear down an already-adopted peer's live session — two branches, one fix

`p2p/libp2p/server.go` — `handleStream` handshake-failure path, and the MaxPeers/duplicate check ordering in both `handleStream` and `dialPeer`

**(a) New in this PR:** the handshake-failure path calls `ClosePeer(remotePeer)` justified by "the per-peer pending guard above rules out a concurrent legitimate stream on the same connection." That comment is factually wrong: `pendingPeers` only tracks in-flight *inbound* handshakes — an already-**adopted** peer has no entry (and `dialPeer` uses `srv.dialing`, never `pendingPeers`). Concrete honest-node scenario: cross-dial at bootstrap — A's `dialPeer` adopts B while B's concurrent dial has a stream sitting in A's `handleStream`; B aborts mid-handshake (its own 5s deadline under load, shutdown, packet loss) → A's handshake read errors → `ClosePeer(B)` closes every connection to B including the just-adopted healthy session. `Protect()` does not defend against an explicit `ClosePeer`. One failed redundant stream becomes loss of a live peer plus backoff churn.

**(b) Pre-existing, but in code this PR modified and worth fixing together:** in both `handleStream` and `dialPeer`, the `MaxPeers` reject (whose branch calls `ClosePeer` under a "we never adopted a stream to this peer" comment) runs *before* the duplicate-peer check (which deliberately does `Reset()` only, with a comment explaining exactly why `ClosePeer` would be wrong). At capacity these two conditions co-occur naturally: the concurrent adoption of the *same peer* is what pushed the map to `MaxPeers`, so the `ClosePeer` branch fires on a peer whose healthy session was just adopted.

Both are the same hazard — calling `ClosePeer` on a `peer.ID` that may have a concurrently-adopted session — and one fix covers both: consult `peerMap` (under `peerMu`) before any pre-adoption `ClosePeer` and `Reset()`-only when the peer is adopted, plus reorder duplicate-before-MaxPeers in both functions. Mitigating factor: streams are noise-authenticated to the remote peer, so this is honest-peer churn and self-inflicted damage, not third-party attacker leverage.

## 5. [P3] Peer-database cap: hourly-only enforcement and pure-recency eviction

`p2p/libp2p/peerdb.go` — `expire` / `recordConnected`

Two related softnesses in the new 2000-record cap, both worth a cheap hardening rather than a redesign:

- **Enforcement cadence:** the cap is enforced only in `expire()` (hourly + at open) — deliberately, per the code comment, to keep `recordConnected` a single get+put. Between passes the DB can exceed the cap. In practice growth is tightly bounded — each record requires a *completed* Zenon handshake and adoption (a `peerMap` slot, connection-manager budget, several round trips), and records are small — so disk exhaustion within the hour isn't realistic. Still, a cheap in-memory insert counter that triggers an early eviction pass when it crosses, say, 2× the cap would close the gap without touching the write path's cost profile.
- **Eviction policy:** when over cap, `expire()` keeps the newest-`lastSeen` records unconditionally. Long-connected honest peers only get `lastSeen` stamped at adoption/disconnect, so their records look stale while an identity-cycling source's records look fresh — a flush of ≥2000 ephemeral identities shortly before the hourly pass evicts genuine long-standing peers rather than just excess growth. To be fair, this is *not* a regression: pre-PR `seeds()` was a pure-recency prefix that was strictly easier to dominate, and this PR's subnet round-robin is a real improvement — but the eviction now decides what *permanently survives*, so it deserves the same diversity treatment. Consider tie-breaking eviction by subnet (reuse `subnetKey`) or tracking a first-seen stamp so tenure counts.

Related note: `seeds()` subnet diversity keys off the peer's *self-advertised* identify addresses, which an identity-cycling source controls — so the round-robin spread is advisory against a deliberate attacker (mitigated by dial-failure counters pruning undialable records).

## Smaller items (non-blocking)

- `p2p/libp2p/server.go` (connmgr setup) — the comment "Bound raw libp2p connections by the same limits MaxPeers/MaxPendingPeers" overstates what `BasicConnMgr` does: it's trim-only (no admission control), protected peers are excluded from trim candidates, and with `MaxPeers` connections protected the `ncandidates < lowWater` early-return means the trimmer essentially never fires — the effective ceiling is roughly `protected + lowWater` (~2× MaxPeers), with the only hard cap being the resource manager's defaults. No functional regression (still far stricter than libp2p's 160/192 defaults, and verified against go-libp2p v0.33.2 that DHT connections are *not* thrashed), but the comment should be rewritten and, if a real cap is wanted, rcmgr `ConnsInbound` limits are the mechanism.
- `verifier/momentum.go` (new early bound) — the bound inlines `config.MaxBasePlasmaInMomentum / constants.EmbeddedSimplePlasma`, duplicating the formula in `dp`'s `MaxContractBlocksInMomentum()` used 50 lines below and in the content selector. The values match exactly today (the bound is arithmetically exact — verified), but a consensus-path cap now lives in two places; a package-level `dp` helper both call would remove the drift risk.
- `chain/account_pool.go` — `higherPriority`'s dp branch does 4–5 LevelDB read sequences per contested insert *under the pool lock inside the chain insert path* (`IsSporkActive` iterates all sporks and deserializes the frontier momentum, `GetFrontierMomentum` re-reads that same momentum, `GetPlasmaVariables` reads embedded storage), though all inputs change only per momentum. The pool is already a registered `InsertMomentum` listener, so the comparator inputs (or a cached `DynamicPlasma`, as the content selector already does) can be refreshed once per momentum. At minimum, fetch the frontier momentum once and reuse it for both the spork check and `NewDynamicPlasma`.
- `chain/account_pool.go` — two behavior notes on the dp branch: (1) `dp.HigherPrice` returns `ErrBlockPriceSame` on exact ties with no hash tie-break, so equal-priced same-height replacements are now first-seen-wins and order-dependent per node (legacy converged all pools on the lower-hash variant; mempool-local only, since momentum application force-adds content). (2) Any error from the three store lookups silently falls back to the legacy comparator with no log line, in a function that logs every other rejection path. Suggest an explicit tie-break on `ErrBlockPriceSame` and a log on fallback.
- `p2p/libp2p/server.go` — `pendingCount` (atomic) and `pendingPeers` (map) now track the same population 1:1 (the counter's window is strictly enclosed by the map's, verified against all uses repo-wide); enforcing the global limit as `len(srv.pendingPeers) >= limit` inside the existing `peerMu` section would leave one source of truth.
- `p2p/libp2p/stream_rw.go` — `ReadMsgLimited` is dead code as of this PR (its only caller moved to `ReadMsgLimitedTimeout`), and its doc comment still advertises it for the handshake path — inviting reuse with the wrong 30s timeout.
- `p2p/libp2p/server.go` — `Start()` normalizes `MaxPeers`/`MaxPendingPeers` fallbacks for the connmgr sizing, `handleStream` duplicates the `MaxPendingPeers` fallback, but the adoption checks use raw `srv.MaxPeers` — so a (documented-invalid) `MaxPeers: 0` config gets a connmgr sized for 70 that admits 0. Normalizing both fields once at the top of `Start()` would make the three sites agree.

## What I verified and agree with

- The per-peer pending-handshake cap, the 5s `handshakeTimeout` via `ReadMsgLimitedTimeout`, and `Protect()` on adoption are correct and address the review comments well; `TestPendingHandshakeCapPerPeer` covers the right scenario.
- Moving `checkUncommittedBlocksCount` into the fast-forward branch is right — replacements shorten the pending chain and duplicates are idempotent (modulo item 3 above).
- The embedded-contract memoization is safe: each of the 16 variants is built from a fresh `getOrigin()` (fresh maps and method structs per iteration), diff functions only touch their own variant's map, and everything is read-only at runtime.
- The `enoughPlasma` `return err` restoration is correct — `DealWithErr` was a remote-input panic path.
- The verifier's local-copy fix (`priced := *block`) is safe (only the scalar `BasePlasma` differs, downstream calls are read-only), and the restored dynamic-plasma content-size pre-bound cannot reject an otherwise-valid momentum (canonical user base plasma ≥ 21,000; contract count capped at `Max/EmbeddedSimplePlasma`).
- The spork-aware pool `higherPriority` correctly aligns replacement ranking with content selection (all three `Stable` implementers updated; no new lock-order cycle — momentum-pool broadcasts run with its lock released).
- `Protect`/`Unprotect` ordering is race-free (`Unprotect` strictly precedes the `peerMap` delete that gates re-adoption).

digitalSloth:
Thanks for the thorough review — went through every point, verified the ones I wasn't sure about against the code, and pushed fixes for everything I agreed was a real issue. Summary below, organized the same way as your review.

## P1s

**Content selector fee-priority escort** — confirmed. `filterBlocksToCommit` has no per-address block cap on the dp path, `DefaultMaxBasePlasmaInMomentum` is exactly 200 × 21,000, and `MaxUncommittedBlocksPerAccount` is 500, so a single premium head could carry a floor-priced tail through the whole momentum budget ahead of a competing address's premium block. That's a regression the group-contiguous ranking introduced — the intransitivity fix shouldn't have cost per-block fee priority.

Fixed with a k-way merge over per-address heads (`container/heap`, pop by price + hash tie-break, promote the successor) instead of emitting whole groups contiguously. This restores per-block price ordering while still guaranteeing ancestor-before-descendant emission by construction, so the no-gap property holds. Embedded-address groups still stay contiguous and go first — `filterBlocksToCommit`'s contract-batch accumulation depends on that. Added a regression test for the exact escort scenario (premium head + floor-priced tail vs. a competing premium block from another address).

**`ValidPrice` saturation** — arithmetic confirmed. Reproduced the vector exactly: at `fusionPrice = 1<<61`, `workPrice = 1000`, a block with only 10,000 PoW plasma scored above a saturated floor that should have required ~21,000. Saturation truncated the floor downward, so the old "fails closed" comment was wrong — it was the opposite.

Fixed by comparing directly in `big.Int` instead of round-tripping through a synthetic block, keeping the `PriceScaleFactor` division ahead of the multiplications so non-overflow results stay bit-identical (this is on the verifier's per-block path). Test extended with your vector (10,000 rejected / 21,000 accepted).

One note on severity, not correctness: I don't think the overflow is actually reachable in practice — it needs `fusionPrice > ~8.8e17`, but a fused-paid block requires `FusedPlasma >= 21 * fusionPrice`, and `MaxFusionPlasmaForAccount` caps out around 2.1e11. Past `fusionPrice ≈ 1e10` no account can pay a fused block at all, so fused usage drops to zero and the price decays instead of climbing — it can't reach the overflow regime. Fixed it anyway since it's a small, strictly-more-correct change.

## P2s

**Removed RPC cap check → silent success for dropped blocks** — confirmed, and the scope is actually broader than the missing check alone: `CreateAccountBlock` returned `void` and only logged, so *every* pool rejection (cap, `canRollback`, `higherPriority`, previous-mismatch) was already being swallowed before this PR, not just the cap case.

Fixed by propagating the insert error properly instead of reinstating a narrow pre-check: `Broadcaster.CreateAccountBlock` now returns an error, which flows through to `PublishRawTransaction` and to the pillar block-production loops. Six call sites touched (interface, broadcaster, RPC, two pillar workers, the mock). Added an RPC test that submits past `MaxUncommittedBlocksPerAccount` and asserts the error comes back. Side benefit: `pillar/worker.go`'s autoreceive loop previously kept regenerating and re-inserting the same failing block forever on a swallowed error — it now aborts instead.

**Pre-adoption `ClosePeer` can tear down an adopted peer's session** — both halves confirmed, and the handshake-failure case was actually worse than "incomplete": `Network().ClosePeer` is peer-scoped and closes *every* connection to that peer, so "the failing stream is the only one on this connection" wouldn't have been a sufficient guard even if `pendingPeers` covered outbound dials — it doesn't, `dialPeer` uses a separate `dialing` map.

Fixed both:
- Handshake-failure path now only closes when the peer has **no adopted session and no dial in flight** (checks both `peerMap` and `dialing` under the same lock before releasing and calling `ClosePeer`).
- The `MaxPeers`-at-capacity branch is now ordered *after* the duplicate-peer check in both `handleStream` and `dialPeer`, so a redundant stream from an already-adopted peer gets `Reset`, not closed.

Added tests for both: an adopted peer survives a failed second handshake, and a duplicate stream at capacity is reset rather than closing the live session.

## P3s (non-blocking)

**Peer-DB cap enforced only hourly** — agreed it's a soft edge, but leaving it. Records are only written by `recordConnected`, which runs after a completed handshake *and* adoption, so growth is bounded by adoption churn against `MaxPeers` already — an insert-side counter would buy a bound the handshake cost provides for free, at the cost of turning a hot-path get+put into more. The temporary overshoot is already documented on the constant.

**Peer-DB eviction is pure recency** — fair point, and slightly different mechanism than framed: `lastSeen` is stamped at adoption and at disconnect, so a peer connected for hours looks *stale* the whole time it's connected. Didn't implement the fix in this pass (needs >2000 successful adoptions inside one cleanup cycle to bite), but the cheap targeted version — refresh `lastSeen` for currently-connected peers during the hourly cleanup pass, before `expire()` runs — is a good follow-up if you'd rather not leave it.

## Smaller items

- **connmgr comment overstating `BasicConnMgr`** — agreed, reworded. `BasicConnMgr` trims, it doesn't refuse, and protected peers are excluded, so the watermarks are a trim target, not a hard bound. Sizing and grace period unchanged.
- **Verifier duplicating the max-content formula** — half-agreed, left as-is. They're not actually the same formula (verifier's bound is user+contract blocks, `MaxContractBlocksInMomentum()` is only the contract term), and the verifier needs the bound before a `DynamicPlasma` instance exists. Would rather not add a third spelling for a partial overlap.
- **dp-aware `higherPriority` store reads** — noted, deferred. That branch only runs on contested replacement, never on fast-forward inserts, so it's a rare path; caching adds invalidation state for not much payoff. Will revisit if it shows up in profiling.
- **`ErrBlockPriceSame` has no hash tie-break** — agreed, and more than a nit: the dp branch returned `HigherPrice` directly, so an exact tie was non-nil and first-seen won — order-dependent across nodes, meaning two nodes seeing the same conflicting blocks in different order wouldn't converge. Fixed with a smallest-hash tie-break, matching the legacy branch and the rule already documented on the `AccountPool` interface. (Note: the content selector's own tie-break goes the *other* way, larger hash — that's ordering rather than conflict resolution, a deliberate difference, not an inconsistency to fix.)
- **Silent legacy-comparator fallback** — agreed, added a debug log so a store-lookup failure is visible instead of silently changing which comparator ran.
- **`pendingCount` duplicating `pendingPeers`** — agreed, collapsed into `len(srv.pendingPeers)` inside the existing locked section, which also makes the per-peer and global checks a single atomic decision.
- **`ReadMsgLimited` dead code** — agreed, deleted. No callers, not part of `MsgReader`/`MsgReadWriter`, and this PR is what orphaned it.
- **Peer-limit fallbacks normalized in three places** — agreed, normalized once in `Start()` and used everywhere. Worth flagging: `MaxPeers == 0` previously meant "accept no peers" (an always-true `>= 0` check); after normalization it gets the same 60-peer preset `MaxPendingPeers` already got. No production path sets it to zero, but noting the behavior change for the record.

Thanks also for calling out what was already correct — saved time not re-litigating those.

All of the above is implemented, tested (including `-race` on the touched p2p and chain packages), and independently reviewed for correctness and security before this reply.

0x3639:
# PR #74 re-review — process, collective findings, and proposed fixes (head `4390b70`)

**Verdict: request changes.** The author's five fix commits (`e7f5336..4390b70`) genuinely resolve every major finding from the [first-round review](https://github.com/zenon-network/go-zenon/pull/74#pullrequestreview-4936577469) — each fix was independently re-verified here, not taken on faith. Three items remain before merge: a race in the p2p `ClosePeer` teardown path (real, confirmed twice, present at three call sites), an error-propagation change that makes the pillar's contract loops abort a whole producer slot on the first failed insert, and a silent config reinterpretation that turns a `MaxPeers: 0` isolation setting into 60 peers at spork switchover. Concrete patches for every actionable item are in §4 and §5.

Context for the reader: this document consolidates **three** review passes. After the fix commits landed, an independent re-review by Codex (OpenAI's coding agent) was run against head `4390b70`; it also concluded "request changes", flagging the `ClosePeer` race as the one remaining blocker. This document adjudicates that re-review — §2 confirms its finding but widens its scope from one call site to three and rejects one of its two proposed fixes — and adds the findings of this session's own re-review (direct verification plus a multi-agent pass), which surfaced the other two blocking items Codex did not report.

---

## 1. How this review was produced

Three independent inputs were adjudicated against each other, with the code at `4390b70` as the only arbiter:

1. **The [first-round review posted on the PR](https://github.com/zenon-network/go-zenon/pull/74#pullrequestreview-4936577469)** (2026-08-14): five findings (P1 content-selector fee escort, P1 `ValidPrice` saturation, P2 RPC silent drop, P2 pre-adoption `ClosePeer`, P3 peer-DB cap) plus eight smaller items, written against the pre-fix head. The PR author responded with five commits (`e7f5336..4390b70`).
2. **Codex's re-review** of `4390b70`: verdict "request changes" — prior majors fixed, one remaining P2 (`ClosePeer` stale adoption snapshot), with a claimed deterministic reproduction and two proposed fix directions.
3. **This session's re-review**, in three passes:
   - **Direct verification** of each fix commit: the k-way merge was read end-to-end and checked for strict-weak-ordering validity; the new `ValidPrice` was checked term-for-term against the old `HigherPrice(block, minBlock)` semantics; the `CreateAccountBlock` interface change was traced through every caller; the `ClosePeer` race window was traced instruction-by-instruction, including what `host.Connect` does on an existing connection and what `Protect()` does and does not defend against.
   - **A multi-agent review** of the full PR diff: 8 finder angles (line-by-line scan, removed-behavior audit, cross-file tracer, efficiency, altitude, and three more) produced 26 candidate findings, which were deduplicated and then passed through an adversarial verification stage that assigned each survivor a CONFIRMED or PLAUSIBLE verdict. Ten survived the cap; the cut and the explicitly-verified-clean list are folded in below.
   - **Validation**: `go test -race -count=1 ./p2p/libp2p ./dp ./pillar ./chain ./protocol ./rpc/api` on a clean worktree of `4390b70` (`GOWORK=off`) — all pass. Like Codex, we observe no CI check runs on the head commit.

Nothing below is included on any reviewer's say-so alone: every finding in §3 was re-checked against the file and line it names, and two of the agent-sourced candidates were re-verified by hand before ranking (the `MaxPeers: 0` semantics flip was traced through `app/config.go` → `node/config.go` → `p2p/switcher/server.go:289` → the new normalization; the connmgr grace premise was checked against the actual `NewConnManager` call at `server.go:531`).

## 2. Adjudication of Codex's re-review

**The remaining P2 is confirmed.** The handshake-failure path snapshots `peerMap`/`dialing` under `peerMu.RLock` (`p2p/libp2p/server.go:802-805`), releases the lock, then calls `Network().ClosePeer` (line 807). A `dialPeer` for the same peer that starts after the snapshot can reserve its `dialing` slot, reuse the same host connection via `host.Connect` (a no-op when already connected), complete its handshake, adopt the peer — and the stale-snapshot `ClosePeer` then closes **every** connection to that peer ID, including the just-adopted session. `ConnManager().Protect()` is no defense: it exempts a connection from the trimmer, not from an explicit `ClosePeer`. There is also a milder variant Codex didn't mention: if the racing dial has started but not yet adopted, the close kills our own in-flight outbound handshake, incrementing the peer-DB `FailCount` (toward fail-out eviction) and doubling the dial backoff.

Two caveats on the report itself:

- **The "deterministic repro" necessarily instrumented the code.** There is no seam between the snapshot and the close in the shipped source — the only test hook (`testDialAdoptHook`) fires inside `dialPeer`, after its handshake. A deterministic reproduction had to pause `handleStream` artificially. That is a legitimate existence proof of the interleaving, but should be reported as such.
- **Production incidence is low.** The window is normally sub-microsecond; hitting it requires the `handleStream` goroutine to be descheduled for a full connect + handshake round trip while our own maintenance loop happens to dial that exact peer (essentially cross-dial at bootstrap). No third party gains leverage — streams are Noise-authenticated, and an attacker can neither schedule our dials nor stall our goroutine. On incidence alone this is P3; it still blocks because the PR's purpose is connection stability and the fix is small.

**The scope is wider than Codex reported.** The same unlock-then-`ClosePeer` shape exists at the MaxPeers-reject branches of both `handleStream` (~lines 893–902) and `dialPeer` (~lines 1093–1101): "the duplicate check already ruled out an adopted peer" is also a snapshot that goes stale at `Unlock`. At capacity, a slot freed by a concurrent disconnect plus an adoption by the opposite path inside the window kills a freshly adopted session identically. The `handleStream` MaxPeers branch additionally never checks `dialing`, so it can kill our own outbound handshake with no adoption race needed. One mechanism must cover all three sites (patch F1 below).

**Of Codex's two proposed fixes, take the reservation, reject reset-only.** Option (b) — reset the stream and let the connection manager reclaim the unadopted connection — does not work in this codebase: `BasicConnMgr` is trim-only, protected peers are excluded from candidacy, and the trimmer early-returns when candidates number fewer than the low watermark, so on a healthy node it essentially never fires (the PR's own rewritten connmgr comment now says as much). Under (b), connections from peers that fail the handshake would linger indefinitely holding a yamux session and file descriptor — precisely the untracked-connection problem the `ClosePeer` calls were added to solve.

All items on Codex's "previous findings fixed" list were independently confirmed — details in §6.

## 3. Collective findings

Sources: **[C]** Codex re-review, **[R1]** first-round review, **[A]** this round's multi-agent pass, **[D]** this round's direct verification. Verdicts are from the adversarial verification stage where one ran.

| # | Sev | Location | Finding | Source | Verdict |
|---|-----|----------|---------|--------|---------|
| F1 | P2 | `p2p/libp2p/server.go:806` (+`:901`, `:1100`) | Pre-adoption `ClosePeer` decided on a stale snapshot; three sites share the pattern | C + A + D | CONFIRMED |
| F2 | P2 | `pillar/worker.go:144`, `pillar/worker_updater.go:31` | Error propagation makes both pillar contract loops abort the whole slot on the first failed insert | A + D | CONFIRMED |
| F3 | P3 | `p2p/libp2p/server.go:463` | `MaxPeers: 0` silently normalized to 60, flipping legacy "no non-trusted peers" isolation at spork switchover | A (verified D) | CONFIRMED |
| F4 | P3 | `p2p/libp2p/server.go:531/914/1113` | connmgr grace = 5s but `Protect()` fires only at adoption; slow dials and just-adopted peers are trim candidates above the high watermark | A (premise verified D) | PLAUSIBLE |
| F5 | P3 | `p2p/libp2p/server.go:762` + `:806` | Dead pending-handshake stream locks out a reconnecting peer for ~5s; timeout's `ClosePeer` then also kills the retry connection | A | PLAUSIBLE |
| F6 | P3 | `chain/account_pool.go:162-176` | `higherPriority` fails open to the legacy comparator on store errors (Debug-only), and decodes the frontier momentum twice per contested insert under both pool locks | A + R1 | PLAUSIBLE / CONFIRMED |
| F7 | P3 | `p2p/libp2p/peerdb.go:145` | `subnetKey` claims peerdiversity parity but buckets /24 vs. the filter's /16 (one attacker /16 = 256 seed groups); DNS-only addrs all collapse to key `""` | A | CONFIRMED |
| F8 | P4 | `dp/dp.go:167`, `verifier/momentum.go:212` | Consensus formulas duplicated: 4th hand-written copy of the price-value expression; verifier inlines the contract-count cap that `dp` provides | A + R1 | CONFIRMED |
| F9 | P4 | `vm/embedded/embedded.go:212` | Hard-coded 16-variant array: a 5th spork bit without doubling it compiles clean and panics network-wide at activation | A | CONFIRMED |
| F10 | P4 | `zenon/mock/zenon.go:231,268` | Mock helpers discard the new `CreateAccountBlock` error, so pool-rejected inserts pass tests silently | A | CONFIRMED |

Cut below the cap after verification (noted, no action requested): at-cap replacement churn in the account pool (amortized O(1), price-escalation-bounded); `handleStream`/`dialPeer` adoption-section duplication; `seeds()` loop-flag complexity; unbatched `expire()` deletes. The hourly peer-DB cap/recency eviction from R1 remains the acknowledged non-blocking follow-up; F7 should fold into it.

## 4. Blocking items — proposed code changes

### F1. Atomic close reservation for all pre-adoption `ClosePeer` sites

Add a `closing` set alongside `dialing`/`pendingPeers` (guarded by `peerMu`, initialized in `Start()` with the other maps), and a helper that re-checks and reserves atomically:

```go
// closePeerIfUnadopted tears down every connection to id unless the peer
// has an adopted session or a dial in flight — re-checked under the
// write lock, atomically with reserving the close, so no dial or
// adoption can begin while the close runs. ownDial is set when the
// caller is the peer's own dialPeer goroutine (whose dialing entry is
// still registered and must not veto the close).
func (srv *Server) closePeerIfUnadopted(id peer.ID, ownDial bool) {
	key := id.String()
	srv.peerMu.Lock()
	_, adopted := srv.peerMap[key]
	_, dialing := srv.dialing[key]
	if adopted || (dialing && !ownDial) {
		srv.peerMu.Unlock()
		return
	}
	srv.closing[key] = struct{}{}
	srv.peerMu.Unlock()

	_ = srv.host.Network().ClosePeer(id)

	srv.peerMu.Lock()
	delete(srv.closing, key)
	srv.peerMu.Unlock()
}
```

Apply it at the three snapshot-justified sites:

- **`handleStream` handshake-failure path** (lines 802–808): replace the RLock snapshot + conditional close with `srv.closePeerIfUnadopted(remotePeer, false)`.
- **`handleStream` MaxPeers branch** (line 901): replace the bare `ClosePeer` with `srv.closePeerIfUnadopted(remotePeer, false)` — this also adds the `dialing` check that branch is currently missing.
- **`dialPeer` MaxPeers branch** (line 1100): replace with `srv.closePeerIfUnadopted(info.ID, true)` (our own `dialing` entry is still registered until the deferred delete, hence `ownDial`; no other dial can coexist because the reservation refuses duplicates, so only inbound adoption needs excluding — which the `adopted` re-check plus the reservation handle).

Then make the reservation visible to the three places that could otherwise race it:

```go
// dialPeer, inside the initial reservation critical section:
	if _, exists := srv.closing[peerKey]; exists {
		srv.peerMu.Unlock()
		return fmt.Errorf("close in progress for %s", info.ID)
	}

// handleStream and dialPeer, in the final adoption critical sections,
// after the duplicate check (treat exactly like duplicate — Reset only,
// the stream's connection is being torn down and retry gets a fresh one):
	if _, exists := srv.closing[remotePeer.String()]; exists {
		srv.peerMu.Unlock()
		s.Reset()
		return
	}
```

The maintenance loop's backoff naturally retries a refused dial. The unconditional `ClosePeer` sites (NodeID mismatch, self, protocol mismatch, shutdown) can stay as they are — a peer failing those checks fails them identically as a dial target, so only the authenticated remote can hurt itself there — though routing them through the helper too costs nothing and makes the invariant uniform. A residual sliver remains inside libp2p itself (a connection established after `ClosePeer` snapshots its list), which is unreachable from this code and self-heals on retry. Holding `peerMu` across `ClosePeer` instead would be deadlock-safe today (no notifiees are registered) but would park the server's hot mutex on network I/O — the reservation is the better trade.

Suggested regression test: reuse the `testDialAdoptHook` pattern — add a `testHandshakeFailHook` invoked between the failure and the close decision, hold `handleStream` there while a dial to the same peer completes and adopts, release, and assert `PeerCount()` stays 1 and `Connectedness` stays `Connected` on both sides.

### F2. Pillar contract loops: skip-and-continue instead of aborting the slot

The interface change (`CreateAccountBlock` returning `error`) was made for the RPC path; the two pillar-side callers should not abort the slot on the first failure. In `pillar/worker.go` (autoreceive loop):

```go
			if err := w.broadcaster.CreateAccountBlock(transaction); err != nil {
				// Skip this contract and keep serving the others. A
				// failed insert does not set `one`, so the outer loop
				// cannot spin regenerating a block that repeatedly
				// fails to insert.
				w.log.Error("unable to insert autoreceive block for contract", "reason", err)
				continue
			}
```

`continue` (instead of `return`) restores per-contract independence, and the loop's termination argument survives: the outer `for` only repeats while some contract sets `one = true`, so a contract whose insert always fails cannot drive another round by itself — which also preserves the fix for the old regenerate-and-fail spin that the current `return` incidentally made.

In `pillar/worker_updater.go` (`updateContracts`):

```go
			} else if err := w.broadcaster.CreateAccountBlock(block); err != nil {
				// A failed insert for one contract's update must not
				// starve the remaining contracts in EmbeddedWUpdate;
				// the next momentum retries this one.
				w.log.Error("unable to insert update block for contract",
					"contract-address", address, "reason", err)
			}
```

All update blocks are sent from the coinbase account and `GenerateFromTemplate` reads the live frontier, so continuing after a failed insert generates the next contract's block on the true frontier — no stale-previous chaining. (The pre-existing `return err` on a `GenerateFromTemplate` failure predates this PR and can stay.)

### F3. `MaxPeers: 0` — decide the semantics explicitly, don't rewrite silently

The legacy backend honors an explicit `MaxPeers: 0` as "reject every non-trusted, non-static connection" (`p2p/legacy/server.go`, `encHandshakeChecks`); an explicit `0` in `config.json` survives `json.Unmarshal` over the defaults, is copied verbatim by the switcher (`p2p/switcher/server.go:289`), and the new normalization in `Start()` rewrites it to 60 — so a deliberately isolated node starts adopting peers at Libp2pSpork activation with no warning. The normalization's goal (fixing connmgr-sized-for-70-admits-0) is right; the method shouldn't be silent reinterpretation of a meaningful legacy config. Preferred fix — fail loudly and let the operator decide:

```go
	// p2p/config.go documents MaxPeers as "must be greater than zero".
	// The legacy backend nonetheless treated an explicit 0 as "accept
	// no non-trusted peers"; this backend has no equivalent mode, so
	// reject the config outright rather than silently reinterpreting
	// an isolation setting as the 60-peer default.
	if srv.MaxPeers <= 0 {
		return fmt.Errorf("Server.MaxPeers must be > 0 (got %d)", srv.MaxPeers)
	}
	if srv.MaxPendingPeers <= 0 {
		srv.MaxPendingPeers = p2p.DefaultMaxPendingPeers
	}
```

(The `MaxPendingPeers` zero-default is genuinely documented, so its normalization stays.) If a startup error at spork height is considered too harsh, the alternative is to honor 0 the way legacy does — skip normalizing `MaxPeers` and size the connmgr with `max(srv.MaxPeers, 1)` — but that should be a deliberate choice recorded in the config docs, not a side effect.

## 5. Non-blocking items — proposed code changes

**F4 — Protect inside the adoption critical section.** `BasicConnMgr.Protect` is a tag-map operation with no callbacks, safe under `peerMu`; moving it up closes the inbound half of the window entirely:

```go
	srv.peerMap[remotePeer.String()] = p
	srv.loopWG.Add(1)
	// Protect before releasing peerMu: connmgr grace equals
	// handshakeTimeout, so a peer adopted near the deadline is already
	// a trim candidate the moment the lock drops.
	srv.host.ConnManager().Protect(remotePeer, zenonPeerProtectTag)
	srv.peerMu.Unlock()
```

(mirror in `dialPeer`; `runPeer`'s Unprotect-before-delete ordering is unchanged). The outbound half — a dial whose connect+handshake exceeds the 5s grace — would need a temporary protect tag around the dial (`Protect` at dial start, `Unprotect` in the deferred cleanup unless adopted); worth doing only if slow-link dial churn shows up in the field.

**F5 — reconnect lockout via stale pending slot.** Bounded at ~5–6s plus the peer's backoff. F1's reservation doesn't address it (different mechanism). Cheapest real fix if it matters: key `pendingPeers` by `(peer.ID, stream)` — or track the pending stream and, when a new stream arrives from the same peer, reset the *old* one and take the slot. Recommend a code comment now, follow-up if observed.

**F6 — `higherPriority`: log the fallback and stop decoding the frontier twice.** Minimum change: raise the three `Debug` fallback lines to `Warn` (a comparator-regime switch mid-operation deserves operator visibility). The structural fix is to stop doing store reads in the comparator at all: the pool already registers as an `InsertMomentum` listener, so refresh a cached comparator once per committed momentum and use it here — the same pattern the content selector already uses:

```go
// on each InsertMomentum notification:
//   if spork active: ap.dpMu.Lock(); ap.dp = dp.NewDynamicPlasma(momentum, vars); ap.dpMu.Unlock()
// higherPriority then reads ap.dp under ap.dpMu and falls back to the
// legacy ratio only when it is nil (pre-activation / genesis).
```

That removes `IsSporkActive`'s internal frontier fetch, the second `GetFrontierMomentum` decode, and `GetPlasmaVariables` from the per-insert path entirely.

**F7 — `subnetKey`.** Fix the comment (the peerdiversity filter uses IPv4 /16, legacy /8, and IPv6-by-ASN — the "/24, same granularity" claim is wrong), consider matching /16 for the anti-eclipse property the mechanism advertises, and give DNS-only records their own group instead of the shared `""`:

```go
	// after the IP loop, before `return ""`:
	for _, a := range addrs {
		for _, p := range []int{ma.P_DNS4, ma.P_DNS6, ma.P_DNSADDR} {
			if host, err := a.ValueForProtocol(p); err == nil && host != "" {
				return "dns:" + host
			}
		}
	}
```

Fold into the already-acknowledged peer-DB follow-up.

**F8 — extract the shared consensus formulas.** In `dp`:

```go
// priceValue is the consensus price numerator used by both ValidPrice
// and HigherPrice: FusedPlasma·workPrice + DifficultyToPlasma·fusionPrice.
func (dp *dynamicPlasma) priceValue(block *nom.AccountBlock) *big.Int {
	v := new(big.Int).Mul(new(big.Int).SetUint64(block.FusedPlasma),
		new(big.Int).SetUint64(dp.workPrice))
	return v.Add(v, new(big.Int).Mul(
		new(big.Int).SetUint64(DifficultyToPlasma(block.Difficulty)),
		new(big.Int).SetUint64(dp.fusionPrice)))
}

// MaxContractBlocks is the package-level form of
// MaxContractBlocksInMomentum, for callers that have the variables but
// no DynamicPlasma instance (verifier's early content bound).
func MaxContractBlocks(config *definition.PlasmaVariables) uint64 {
	return config.MaxBasePlasmaInMomentum / constants.EmbeddedSimplePlasma
}
```

`ValidPrice` and both halves of `HigherPrice` call `priceValue`; the method `MaxContractBlocksInMomentum` delegates to `MaxContractBlocks`; and `verifier/momentum.go:212` becomes

```go
	maxContent := config.MaxBasePlasmaInMomentum/constants.AccountBlockBasePlasma +
		dp.MaxContractBlocks(config)
```

so the consensus-path cap has one definition.

**F9 — derive the variant-array size.** One line removes the panic-at-activation trap:

```go
	// 2^(number of spork bits): adding a bit above automatically grows
	// the table; forgetting would otherwise compile clean and panic
	// index-out-of-range on every node at the new spork's activation.
	embeddedVariantCombinations = dynamicPlasmaSporkBit << 1
```

(keep the constant adjacent to the bit list so the "highest bit" coupling is visible).

**F10 — mock should fail the test on a rejected insert.** At both call sites in `zenon/mock/zenon.go` (`InsertSendBlock`, `InsertReceiveBlock`):

```go
		common.FailIfErr(zenon.t, zenon.CreateAccountBlock(transaction))
```

so a pool rejection surfaces as a test failure instead of a log line.

## 6. Codex's "previous findings fixed" list — independently confirmed

Each verified directly against `4390b70`:

- **Content selector**: the k-way merge over per-address heads is exactly the structure R1 asked for — pop the best head, the successor competes on its own price (`heap.Fix` after `pos++`); embedded groups stay contiguous first (which `filterBlocksToCommit`'s batch accumulation genuinely requires); `groupHeap.Less` is a valid strict weak ordering; the fee-escort scenario has a pinning regression test.
- **`ValidPrice`**: the direct big.Int comparison, term-for-term equivalent to the old `HigherPrice(block, minBlock)` semantics including division-before-multiplication order, diverging only where the old uint64 code wrapped. No synthetic saturated block remains.
- **RPC propagation**: `Broadcaster.CreateAccountBlock` returns the insert error through the interface, `PublishRawTransaction` surfaces it, and `TestRPCLedger_PublishRawTransactionReportsPoolRejection` pins it (modulo F2/F10 above).
- **Pre-adoption `ClosePeer` for already-present entries**: adopted/dialing check, duplicate-before-MaxPeers ordering in both functions, and the two new tests cover the adopted-at-snapshot cases; only F1's window remains.
- **Tie-break**: `higherPricedBlock` resolves exact price ties by smallest hash, restoring cross-node mempool convergence, deliberately independent of the selector's emission tie-break.
- **Bookkeeping**: `pendingCount` gone (single source of truth under `peerMu`), `ReadMsgLimited` deleted, limit fallbacks normalized once in `Start()` (see F3 for the semantics cost), connmgr comment now honestly describes trim-only behavior.

Also explicitly verified clean by the multi-agent pass: the verifier's early bound cannot reject an otherwise-valid momentum; the cap-check narrowing is sound (rollback cannot lengthen the pending chain); the embedded variant maps are safe to share; the 5s handshake timeout matches the legacy backend; pool/selector tie-break directions are coherent; the `Broadcaster` interface change reaches every implementer.

## 7. Validation

On `4390b70` (clean worktree, `GOWORK=off`): `go test -race -count=1 ./p2p/libp2p ./dp ./pillar ./chain ./protocol ./rpc/api` — all pass. No CI check runs exist on the head commit.

## Summary of requested changes

1. **F1** — replace the three snapshot-justified `ClosePeer` sites with the atomic close-reservation helper, checked by the dial reservation and both adoption sections (§4).
2. **F2** — make both pillar contract loops skip-and-continue on a failed insert (§4).
3. **F3** — reject (or explicitly honor) `MaxPeers <= 0` instead of silently normalizing to 60 (§4).

F4–F10 can ride along or follow up; none should hold the PR beyond the three above.

*Not posted to GitHub — local review document only.*

digitalSloth:
Thanks for the deep re-review — all ten items are fixed, each independently verified against the code before and after, not taken on faith. Blocking items landed first (`ab08420`, now on `feature/fix-dp-libp2p`); the seven non-blocking ones followed on top (`f0061a3` + `5c703e4`). Summary below in the same order as the review.

## Blocking items

**F1 — pre-adoption `ClosePeer` race (three sites)** — confirmed, and I went with a different fix than either of the two options on the table. Rather than a reservation-map (`closing` set) layered on top of the existing `dialing`/`pendingPeers` bookkeeping, the decision and the `ClosePeer` call now just share one `peerMu` critical section at all three sites — two are a one-line move of `ClosePeer` above the existing `Unlock()`, the third is an `RLock`-snapshot converted to `Lock`+check+close+`Unlock`. No new maps, no sentinel error, no `ownDial` parameter.

That only works if holding `peerMu` across `ClosePeer` can't deadlock or stall the network, so I traced it against the actual pinned deps rather than assuming: no `Notify`/`Notifiee` is registered anywhere under `p2p/`, so nothing under `ClosePeer` re-enters `peerMu` synchronously; `swarm.Conn.doClose` does its teardown locally and dispatches `Disconnected` on a separate goroutine; yamux's `Session.Close` closes the FD *before* waiting on its read/write loops, so the wait is bounded by a local close, not remote-controlled I/O. A peer that goes silent or vanishes can't extend the hold. Concurrency is additionally capped by the existing `pendingPeers`/`MaxPendingPeers` and per-peer handshake machinery, so this doesn't open a new DoS surface — a dedicated security pass confirmed the same.

Two regression tests pin the fix: `TestFailedHandshakeClosesUnadoptedPeerUnderLock` and `TestFailedHandshakeSkipsCloseWhileDialInFlight`, both passing under `-race` (`-count=8`, no flakes).

**F2 — pillar contract loops aborting the whole round on the first failed insert** — confirmed, and broader than just the RPC path: `pillar/worker.go`'s autoreceive loop and `pillar/worker_updater.go`'s `updateContracts` both `return`ed on the first `CreateAccountBlock` error, abandoning every other contract/address in that round. Both now log and move on to the next one instead. In `worker.go` the outer `for { one := false; ... }` loop's termination still holds — a failed insert doesn't set `one = true`, so a permanently-failing contract can't drive the loop into spinning; it just costs one extra failed attempt per pass.

Added a fault-injection hook (`failInsertFor`) to the mock harness and `TestProducerContinuesAfterFailedContractInsert`, which drives the real `worker.work()` path via `InsertNewMomentum` and confirms the round completes past an injected failure and still reaches the other contracts plus `updateContracts`.

**F3 — `MaxPeers: 0` silently reinterpreted as 60** — confirmed, and the initial fix (rejecting it in `libp2p.Server.Start()`) turned out to be in the wrong place: production startup goes through `switcher.Server`, which runs the *legacy* backend first — and legacy happily accepts `MaxPeers: 0` as an isolation mode. A libp2p-only check wouldn't fail at boot, it'd fail at spork-activation swap time, which is worse than the bug (a misconfigured node loses its listener entirely mid-activation instead of just getting 60 peers). Fixed by validating `MaxPeers > 0` in `switcher.Server.Start()`, before either backend starts — a bad config is now rejected at process boot on every real startup path. Kept the `libp2p.Server.Start()` check too, as defense-in-depth.

Worth flagging for the record: an explicit `MaxPeers: 0` in `config.json` that previously ran fine (via legacy) will now fail to start. That's the intended behavior — this backend has no equivalent isolation mode, so failing loudly beats silently reinterpreting — but it's an operator-visible change at upgrade time.

## Non-blocking items

**F4 — connmgr grace window vs. late `Protect()`** — confirmed. `ConnManager().Protect()` was called after `peerMu.Unlock()`, so a peer adopted near the end of the 5s grace period was briefly a trim candidate before it got tagged protected. Moved `Protect()` inside the same locked adoption section in both `handleStream` and `dialPeer` — `BasicConnMgr.Protect` is a plain tag-map insert with no callbacks or I/O, confirmed against the actual libp2p source, so holding `peerMu` across it adds no new blocking surface (independently re-checked by a security pass).

**F5 — dead pending-handshake stream locks out a reconnecting peer for ~5s** — confirmed as a real but bounded limitation. This one only needed a comment, not a behavior change: fixing it properly means keying the pending-slot map by `(peer.ID, stream)` or letting a new stream displace a dead one, which is more surface than this cleanup pass warrants. Documented at the rejection site instead.

**F6 — `higherPriority` fails open silently and double-decodes the frontier momentum** — confirmed. Two changes: the three fallback-to-legacy-comparator log lines went from `Debug` to `Warn`, since a comparator-regime switch mid-operation deserves visibility; and the per-insert `IsSporkActive`/`GetFrontierMomentum`/`GetPlasmaVariables` store reads are gone, replaced by a `*dp.DynamicPlasma` cached on the account pool and refreshed once per committed momentum via its existing `InsertMomentum`/`DeleteMomentum` hooks, guarded by a dedicated mutex. Traced the lock ordering through `momentumEventManager.changes → momentumPool.changes` to confirm no new deadlock or stale-read window — both momentum broadcasts already run with `momentumPool.changes` released, so the refresh is safe. Primed once at chain `Init()` so a restarted post-activation node doesn't run on the legacy comparator until the next momentum. New test (`TestAccountPool_higherPriorityUsesCachedDynamicPlasma`) pins the read side with deterministic literal hashes; a second test (`TestAccountPool_MomentumEventsRefreshDynamicPlasma`) proves the refresh wiring itself is exercised — verified by deliberately removing both refresh call sites and confirming the test fails, then restoring them.

**F7 — `subnetKey` comment overstated parity and DNS-only peers collapsed to one group** — confirmed, and this one needed more care than the review's suggested patch: IPv4 bucketing widened from /24 to /16 to genuinely match libp2p's own peerdiversity granularity (the old comment claimed parity it didn't have), and the doc comment now states the actual bucketing instead of a false claim. For DNS-only peers, the review's own literal suggestion — keying by self-advertised hostname — would have been a step backward: `recordPeerConnected` stores identify-supplied addresses, which are attacker-controlled, so a hostname key lets one attacker mint an unlimited number of seed-diversity groups instead of the one shared group DNS-only peers get today. Went with a single constant `"dns"` group instead — DNS-only peers stop being merged into the address-less `""` group (the original finding) without opening a new Sybil vector. Caught and fixed during this session's own spec review before it reached code.

**F8 — consensus price/cap formulas duplicated across files** — confirmed. Extracted a shared `priceValue` helper in `dp/dp.go` that both `ValidPrice` and `HigherPrice` call instead of each carrying its own `big.Int` arithmetic, and a package-level `dp.MaxContractBlocks(config)` that `verifier/momentum.go`'s early content-size bound now calls instead of inlining the formula — kept as a package-level function rather than a method since the verifier needs the bound before a `DynamicPlasma` instance exists.

**F9 — hardcoded embedded-contract variant array size** — confirmed, and the review's suggested `dynamicPlasmaSporkBit << 1` turned out not to actually derive anything (a new spork bit added elsewhere in the const block wouldn't move that expression). Switched the spork-bit constants to `iota` instead, so `embeddedVariantCombinations` is genuinely `1 << (number of bits)` and grows automatically when a bit is added — confirmed by compiling the const block standalone.

**F10 — mock test helpers silently discarded `CreateAccountBlock`'s error** — confirmed, and broader than the one call site the review flagged: `InsertSendBlock`/`InsertReceiveBlock` now fail the test on an unexpected insert error, and a new `InsertSendBlockRejected(expectedError)` helper was added for the tests that deliberately exercise a pool rejection. Tracing the blast radius turned up a second test relying on the old silent-swallow behavior beyond the one in `vm/embedded/tests` — `TestAccountPool_Priority` in `chain/tests/account_pool_test.go` also submits a same-height replacement expected to be rejected — so both were updated, using `errors.Is` against the correct sentinel for each (one wrapped, one not).

## Verification

`go build ./...` clean across both stages; `go test -race -count=1 ./p2p/libp2p/... ./p2p/switcher/...` and `go test -count=1 ./chain/... ./chain/tests/... ./dp/... ./verifier/... ./pillar/... ./zenon/... ./vm/... ./rpc/... ./node/...` all pass. `TestStateProducer`'s golden pillar log — the only test in the repo asserting on exact `PillarLogger` output — is unaffected by any of the added log lines. Each stage went through an independent spec critique before implementation (both caught real issues — a swap-timing bug in F3 and a Sybil-bypass regression in F7 were fixed before either reached code) and an independent correctness + security review after (the F1-F10 pass caught a backwards doc comment and a test-coverage gap in F6's refresh wiring, both fixed and re-verified). All reviews landed at SHIP.

0x3639:
Nice job sir!!

# PR #74 approval — fix: libp2p & Dynamic Plasma (head `5c703e4`)

**Verdict: approve.** All ten findings from the [re-review](https://github.com/zenon-network/go-zenon/pull/74#issuecomment-5322406656) are genuinely fixed in `ab08420` + `f0061a3` + `5c703e4`. Every claim in your response checked out against the code — and in the three places where you pushed back on the review's suggested patches, you were right and your alternative is better. Details below; nothing here blocks merge.

## Blocking items — verified fixed

**F1 — pre-adoption `ClosePeer` race.** Your fix (one `peerMu` critical section around check + close at all three sites) is simpler than the reservation-set the review proposed, and it's sound. I didn't take the boundedness trace on faith — I verified all three legs against the pinned dependencies:

- go-yamux v4.0.1 `Session.Close()` closes the FD (session.go:300) *before* waiting on `recvDoneCh`/`sendDoneCh`, so the wait is bounded by local teardown, not remote-controlled I/O;
- go-libp2p v0.33.2 `swarm.Conn.doClose` dispatches `Disconnected` notifications on a separate goroutine;
- no `network.Notifiee` is registered anywhere in the repo, so nothing under `ClosePeer` can re-enter `peerMu` synchronously.

The review had noted this option was deadlock-safe but steered away from parking a hot mutex on network I/O — your trace shows the hold is bounded by a local FD close, which removes that objection. No new state, no sentinel error, no extra parameter: better engineering than the suggested patch. `TestFailedHandshakeClosesUnadoptedPeerUnderLock` and `TestFailedHandshakeSkipsCloseWhileDialInFlight` deterministically pin exactly the window from the original finding — the second one holding a dial at the adoption hook while an inbound handshake fails is precisely the interleaving that mattered.

**F2 — pillar loops aborting the round.** Both loops now log and `continue`; the autoreceive loop's termination argument is preserved (a failed insert never sets `one`, so a permanently-failing contract can't spin the outer loop), and the incidental fix for the old regenerate-and-fail spin survives. The `failInsertFor` fault-injection hook plus `TestProducerContinuesAfterFailedContractInsert` driving the real `worker.work()` path is exactly the right test shape.

**F3 — `MaxPeers: 0`.** Good catch on the review's suggested placement being wrong: a libp2p-only `Start()` check would have fired at spork-swap time, after the legacy backend was already down — a misconfigured node losing its listener mid-activation is worse than the bug. Validating in `switcher.Server.Start()` before either backend starts, with the libp2p check kept as defense-in-depth for directly-embedded servers, is the right structure. Flagging the operator-visible upgrade change in your response was also the right call — that note should probably make it into the release notes when this ships.

## Non-blocking items — verified fixed

- **F4** — `Protect()` now inside the adoption critical section in both paths; confirmed `BasicConnMgr.Protect` is a tag-map insert with no callbacks, so no new blocking surface under `peerMu`.
- **F5** — documented-only is the right scope; the comment at the rejection site says exactly what the review would have wanted a future maintainer to know.
- **F6** — the cached `DynamicPlasma` with a leaf `plasmaMu`, Warn-level fallback logging, refresh on both `InsertMomentum` and `DeleteMomentum`, and priming at chain `Init()` covers everything the finding asked for, including the restart case. The mutation test on the refresh wiring (delete both call sites, watch the test fail, restore) is appreciated rigor.
- **F7** — you were right to reject the review's DNS-hostname suggestion: keying by a self-advertised hostname would have let one source mint a fresh seed-diversity group per cycled identity — a step backward from the shared group. The single constant `"dns"` group fixes the original finding (DNS-only peers merged into the address-less `""` group) without opening that vector, and the /16 widening makes the comment's parity claim true instead of aspirational.
- **F8** — `priceValue` shared by `ValidPrice` and both halves of `HigherPrice`, and `dp.MaxContractBlocks(config)` replacing the verifier's inlined formula. The consensus-drift risk the finding was about is gone.
- **F9** — correct critique of the review's `dynamicPlasmaSporkBit << 1` (it didn't actually derive anything); the `iota` construction genuinely grows the table when a bit is added above the last line.
- **F10** — `FailIfErr` at both mock call sites, the `InsertSendBlockRejected` helper with `errors.Is` (right choice given the mixed wrapped/unwrapped sentinels), and both dependent tests updated — including the second one in `chain/tests` beyond the one the review flagged.

## Independent verification

On `5c703e4` in a clean worktree: `go build ./...` clean; `go test -race -count=1 ./p2p/libp2p ./p2p/switcher ./chain ./pillar ./dp` and `go test -count=1 ./chain/tests ./verifier ./zenon/... ./vm/embedded/tests ./rpc/api ./protocol` — all pass.

## Two residual notes (no action needed for this PR)

1. The MaxPeers-reject branches still don't check `dialing`, so at capacity an inbound rejection can close a connection carrying our own in-flight outbound dial to the same peer. At capacity that dial was doomed to the same rejection anyway, and backoff self-heals the rare case where a slot freed in between — worth a comment if either branch is ever touched again, nothing more.
2. `DeleteMomentum`'s refresh reads the frontier at listener-fire time, so in a rollback burst the cached pricing context could lag by one momentum until the next event fires. Self-correcting and harmless at current momentum cadence.

Along with the previously acknowledged peer-DB cap/recency follow-up, these can live as follow-up issues if anyone wants to track them; none affect this merge.

Thanks for the quality of the responses across all three rounds — verifying each finding independently before fixing it, and pushing back with evidence where the review's suggestions were wrong, is exactly what this process should look like.

0x3639:
I created three issues (https://github.com/zenon-network/go-zenon/issues/76, https://github.com/zenon-network/go-zenon/issues/75, https://github.com/zenon-network/go-zenon/issues/77) to track the residual note for consideration later

PR #80, opened 2026-09-02
ci: test action and build workflows

digitalSloth:
## Summary

- Add a dedicated `Test` workflow that builds and runs the full test suite on PRs targeting `dev`, `testnet`, and `master`, with output kept quiet on success (`-v` dropped so only failures print detail), a read-only `contents` token, and a concurrency group that cancels superseded runs on force-push.
- Remove `p2p/legacy/nat/natupnp_test.go`'s `TestUPNP_DDWRT`: it hangs waiting on a multicast UDP response with no timeout in sandboxed/CI network environments, and nothing else in the package depends on its `fakeIGD` helper.
- Extend the release workflow (`znn_builder.yml`) to build pre-releases from `dev` and `testnet` pushes as well as `master`, tagging and naming each branch's release distinctly (`-dev` / `-testnet` vs `-alphanet`) and marking `dev`/`testnet` as GitHub pre-releases.
- Move master releases to a `v*` git-tag trigger instead of every push to `master`, so cutting an official release is an explicit `git tag && git push --tags` rather than automatic on merge.
- Replace the hardcoded `metadata.Version`/`metadata.GitCommit` constants with build-time `-ldflags -X` injection: `dev`/`testnet` builds derive their version from `git describe` (unique per commit, no manual bump needed), tagged builds use the tag verbatim, and the commit hash comes from `github.sha` (previously always empty on release binaries, since `buildvcs: false` on the xgo build steps silently broke the old `debug.ReadBuildInfo()` lookup).
- Enable `govulncheck` in the release workflow (previously present but commented out and non-functional). Currently `continue-on-error: true`, since the repo has 25 known reachable vulnerabilities today (mainly an outdated `go-ethereum`) that need dependency bumps, not a CI change, to clear — flip it to blocking once those land.

## Test plan

- [x] `Test` workflow passes on a PR against `dev`
- [x] Push to `dev` produces a `-dev` pre-release with a version derived from `git describe` and a populated commit hash
- [x] Tagging a commit `vX.Y.Z(-alphanet)` and pushing the tag produces a non-prerelease build using that tag as the version
- [ ] `znnd version` / RPC `stats` report the injected version and commit hash on a release binary


edgepillar:
Thanks for adding the PR test workflow and explicit release versioning. I reviewed `96438d16303d8f75d8d6bca7bb9e9da57f7df7bc` and found a version-selection issue worth correcting before the release workflow is enabled.

This is a source-level review; I have not executed the release workflow or a tag-history reproducer.

**[P2] Exclude generated prerelease tags from the base-tag search**

The [base-tag lookup](https://github.com/zenon-network/go-zenon/blob/96438d16303d8f75d8d6bca7bb9e9da57f7df7bc/.github/workflows/znn_builder.yml#L43-L54) excludes `*-dev` and `*-testnet`, but the tags it generates have the forms `vX.Y.Z-dev-N-gHASH` and `vX.Y.Z-testnet-N-gHASH`. Those names do not match either exclusion.

Once a generated prerelease tag is reachable from a later build, it remains eligible for `git describe`. If selected, `BASE_VERSION` retains the previous prerelease suffix and another one is appended, producing a name such as `vX.Y.Z-dev-N-gOLD-dev-1-gNEW`. The same gap permits a reachable tag from the other prerelease channel to become the base.

Could the lookup select only the intended release-base tags, or exclude the actual generated families as well, such as `*-dev-*` and `*-testnet-*`, while preserving the intended handling of legacy tags?

Focused regression cases would be:

1. Two consecutive dev releases after a stable/alphanet base tag.
2. Two consecutive testnet releases.
3. Both prerelease tag families present in the reachable history.
4. An explicit stable/alphanet tag build, which should still use its tag verbatim.

The assertions should pin the selected base, the count from that base, a single channel suffix, and the prerelease flag. These checks should exercise the version calculation without publishing a release or requiring a cross-platform build.

digitalSloth:
Confirmed and fixed — you're right, thanks for the catch.

You'd correctly guessed the intent (dev/testnet builds should always base off the latest real release tag), but the `--exclude` globs only matched the old `vX.Y.Z-dev`/`-testnet` tag shape, not the `vX.Y.Z-dev-N-gHASH` shape this workflow actually generates. Since `svenstaro/upload-release-action` creates a real, reachable tag for every release, a later build on the same branch picked up the prior prerelease tag as its base and compounded the suffix — reproduced locally:

```
v0.0.8-dev-5-gOLD-dev-1-gNEW
```

Fix adds `--exclude '*-dev-*'` and `--exclude '*-testnet-*'` (keeping the old exact-suffix excludes for any legacy tags), so the lookup falls back to the real base tag correctly. I also pulled the version calculation out of the inline `run:` block into `.github/scripts/determine-version.sh` and added `.github/scripts/determine-version.test.sh`, covering exactly the four cases you listed — two consecutive dev releases, two consecutive testnet releases, both prerelease families reachable, and an explicit tag build — each pinning the selected base, count, single channel suffix, and prerelease flag. It's wired into the PR test workflow so it runs on every PR without needing a real release or cross-platform build. Verified it fails against the old globs and passes against the fix.

0x3639 (commented):
Reviewed commit `ee66bb59c80947e8c951416a1ca8fea62be62a02`. I found two issues to address before merging.

1. **[P2] Create prerelease tags at the build commit**

   In [[znn_builder.yml](https://github.com/zenon-network/go-zenon/blob/ee66bb59c80947e8c951416a1ca8fea62be62a02/.github/workflows/znn_builder.yml#L173-L177)](https://github.com/zenon-network/go-zenon/blob/ee66bb59c80947e8c951416a1ca8fea62be62a02/.github/workflows/znn_builder.yml#L173-L177), the release action omits `target_commit`, so newly created tags default to the repository’s default branch, `master`. With the new `dev`/`testnet` triggers, the tag and source archives can therefore reference different code from the published binaries.

   This already occurred in the test release: `v0.0.9-dev-113-gaeaa058` points to `667a69d`, not `aeaa058`. Please set `target_commit: ${{ github.sha }}` so the release tag matches the commit used to build its binaries.

2. **[P2] Preserve commit metadata for source builds**

   Replacing the `ReadBuildInfo` fallback in [[metadata/git_commit.go](https://github.com/zenon-network/go-zenon/blob/ee66bb59c80947e8c951416a1ca8fea62be62a02/metadata/git_commit.go#L3)](https://github.com/zenon-network/go-zenon/blob/ee66bb59c80947e8c951416a1ca8fea62be62a02/metadata/git_commit.go#L3) leaves `GitCommit` empty for the documented `make znnd` build and for `make libznn`, because neither Makefile target supplies the new linker flags.

   A native build reproduced `znnd version` printing a blank commit hash despite the binary containing the correct `vcs.revision`. `stats.processInfo` likewise returns an empty commit. Please retain a fallback when no value is injected, or update the supported build commands to inject it.

Validation: all 16 version-script assertions and the focused NAT tests passed. Native builds succeeded, and linker-injected version/commit output was correct. The latest PR test workflow is green. The earlier prerelease-suffix finding is fixed.

digitalSloth:
Both confirmed and fixed in `43d5b38`.

**1. Tag pointing at the wrong commit** — confirmed by checking the actual test release: `v0.0.9-dev-113-gaeaa058` points to `667a69d` (master's tip at the time), not `aeaa058`. `svenstaro/upload-release-action`'s `target_commit` input defaults to the repo's default branch when the tag doesn't already exist — added `target_commit: ${{ github.sha }}` to both upload steps (package + makefat).

**2. Empty commit hash on native builds** — confirmed, `make znnd`/`make libznn` don't pass any `-ldflags`, so `metadata.GitCommit` stays empty. I went with your first option (retain a fallback) rather than updating the Makefile: restoring the old `debug.ReadBuildInfo()` fallback as a plain var-initializer wouldn't actually coexist with `-X` injection, since Go runs init-time computed initializers unconditionally and that would clobber whatever `-X` set in the CI build. Instead added `metadata.CommitHash()` — returns the injected value if present, else falls back to `vcs.revision` — and switched all three call sites (`app/cli.go`, `app/action_version.go`, `rpc/api/stats.go`) to use it. Verified both paths directly: a plain `go build` resolves to the real HEAD commit via the fallback, and an `-X`-injected build keeps the injected value unclobbered.

openmac0815:
**Review — APPROVE**

The `determine-version.sh` scheme (tag → release version; branch → base tag + commit count + short hash, `-dev`/`-testnet` prerelease markers) with a self-test script is a solid cleanup, and `CommitHash()` falling back to the toolchain-embedded VCS stamp fixes the empty-commit-hash build case. `metadata.Version` becoming an ldflags-injected `var` (default "dev") is the right shape for CI-driven releases.

Conflict warning: this and #101 both edit `metadata/version.go` and `znn_builder.yml`. #101 carries the v0.0.9 community-spork renewal with a wall-clock deadline (momentum 14,455,739, ~2026-10-19) — merge #101 first, then rebase this on top.

0x3639 (commented):
Reviewed commit `43d5b38`. Both items from my previous pass are resolved, thanks.

**How this review was produced.** I ran a Claude Code review over the full `upstream/dev...43d5b38` diff (11 files), then sent those findings to OpenAI Codex (Daybreak, extra-high reasoning) with the same range and asked it to confirm, dispute, or downgrade each one with evidence, plus report anything missed. Codex confirmed six, downgraded four to nits, corrected one scenario, and added four findings. I then ran the checks its sandbox could not. Below is what survived.

**Verified at head** with go1.25.4 and `GOWORK=off`: `.github/scripts/determine-version.test.sh` 16/16 pass, `go build ./...`, `go vet ./...`, and `go test -failfast -p 1 ./...` all clean. The remaining lint/vet output is in files this PR does not touch.

---

**[P1] Test loops pass when `go list` fails**

[test.yml#L41](https://github.com/zenon-network/go-zenon/blob/43d5b3867ec074a39367ea3bab102135641dc75a/.github/workflows/test.yml#L41) and [znn_builder.yml#L54](https://github.com/zenon-network/go-zenon/blob/43d5b3867ec074a39367ea3bab102135641dc75a/.github/workflows/znn_builder.yml#L54) use `for s in $(go list ./...)`. Under GitHub's default `bash -e -o pipefail`, a failing command substitution yields an empty loop and the step succeeds:

```
$ bash -e -o pipefail -c 'for s in $(false); do :; done; echo LOOP_SURVIVED'
LOOP_SURVIVED
```

The PR workflow's earlier `go build ./...` catches most of this, but the release workflow has no equivalent step, so a broken test-only or unused package could ship with no tests run. A plain `go test -failfast -p 1 ./...` avoids the problem.

**[P1] Native builds now report version `dev`**

[version.go#L5](https://github.com/zenon-network/go-zenon/blob/43d5b3867ec074a39367ea3bab102135641dc75a/metadata/version.go#L5) becomes `var Version = "dev"`, but [`make znnd`](https://github.com/zenon-network/go-zenon/blob/43d5b3867ec074a39367ea3bab102135641dc75a/Makefile#L34-L35) and the libznn target pass no `-ldflags`. So `znnd version`, `--version`, `stats.processInfo.version`, and the p2p handshake name at [node/config.go#L233](https://github.com/zenon-network/go-zenon/blob/43d5b3867ec074a39367ea3bab102135641dc75a/node/config.go#L233) all regress to `dev` / `dev-libznn` for anyone building from source. The Makefile already computes `GIT_COMMIT` and never uses it. Either inject `-X .../metadata.Version=$(git describe --tags --always)` and `GitCommit` from the Makefile, or add a `debug.ReadBuildInfo` fallback like `CommitHash()` has.

Related: [`metadata.GitCommit`](https://github.com/zenon-network/go-zenon/blob/43d5b3867ec074a39367ea3bab102135641dc75a/metadata/git_commit.go#L7) stays exported but is now empty on native builds. Anything importing it gets `""` with no compile error. Unexport it, or initialize it from the same fallback so both reads agree.

**[P2] Manual dispatch mints a tag that poisons later dev/testnet versions**

The `*` fallback at [determine-version.sh#L30-L32](https://github.com/zenon-network/go-zenon/blob/43d5b3867ec074a39367ea3bab102135641dc75a/.github/scripts/determine-version.sh#L30-L32) produces `vX.Y.Z-N-gHASH` with `prerelease=false`, and the upload action creates that tag. It matches neither `--exclude` glob, so the next dev push bases off it and yields `v0.0.9-N-gHASH-dev-1-g...`, the compounding bug this PR fixed for the channel tags. Running the script at head with `REF_NAME=master` gives `v0.0.9-100-g43d5b38` / `prerelease=false`. No test case covers this branch. Non-channel dispatches probably should not create a tag at all.

**[P2] Every human-pushed `v*` tag is published as a stable release**

[determine-version.sh#L15](https://github.com/zenon-network/go-zenon/blob/43d5b3867ec074a39367ea3bab102135641dc75a/.github/scripts/determine-version.sh#L15) hardcodes `prerelease=false` in the tag branch, so a pushed `v0.1.0-rc1`, or a `-dev-` / `-testnet-` tag someone pushes by hand, is released and marked Latest. Tags the workflow creates with `GITHUB_TOKEN` do not re-trigger it, so this only bites manual pushes, but it is still one `git push --tags` away. Classify `-rc`, `-dev-`, and `-testnet-` names as prerelease there.

**[P2] Vulnerability job cannot block a release, and runs unpinned code with checkout credentials**

[znn_builder.yml#L63](https://github.com/zenon-network/go-zenon/blob/43d5b3867ec074a39367ea3bab102135641dc75a/.github/workflows/znn_builder.yml#L63) sets `continue-on-error: true`, so a govulncheck failure (or the scanner failing to install) never stops build or publish. [Line 65](https://github.com/zenon-network/go-zenon/blob/43d5b3867ec074a39367ea3bab102135641dc75a/.github/workflows/znn_builder.yml#L65) installs `govulncheck@latest` and runs it in a job whose checkout persisted the token, in a workflow with no `permissions` block. Suggest pinning the govulncheck version, `persist-credentials: false` on that checkout, defaulting the workflow to `contents: read` and granting `write` only to the publishing jobs. If known findings must be tolerated for now, a reviewed baseline is safer than `continue-on-error`.

**[P2] Deleted UPnP test was the only discovery/parser coverage**

`p2p/legacy/nat/natupnp_test.go` was the only test exercising SSDP discovery, device XML parsing, and IGD service matching. The commit message says discovery has no timeout, but vendored goupnp v1.3.0 bounds discovery at 2s and XML fetch at 3s. The CI failure was the `if os.Getenv("CI") != "" { t.Fatalf(...) }` branch on runners without multicast loopback. Dropping that branch, or gating on `testing.Short()`, keeps the test and fixes CI.

**[P3] Pre-existing, but worth fixing while these jobs are being reworked: universal Darwin `znnd` ships non-executable**

`darwin-bins` is [uploaded](https://github.com/zenon-network/go-zenon/blob/43d5b3867ec074a39367ea3bab102135641dc75a/.github/workflows/znn_builder.yml#L124) before the `chmod +x` step, `upload-artifact` strips file modes, and `makefat` runs [`lipo` and `tar`](https://github.com/zenon-network/go-zenon/blob/43d5b3867ec074a39367ea3bab102135641dc75a/.github/workflows/znn_builder.yml#L200-L214) with no `chmod`. The package job repairs its own copy; makefat does not. This is the same on `dev` today, so it is not a regression from this PR.

**[P3] Toolchain drift**

test.yml resolves Go from go.mod (1.23.x), the builder's `test` and `vulnerability` jobs use whatever the runner has preinstalled, and xgo ships 1.24.1. Pinning one version via `setup-go` would make the tests and vulnerability scan exercise the compiler that actually ships.

---

**Nits** (Codex downgraded these and I agree): tag builds keep `-alphanet` verbatim while branch builds strip it; the version test script inherits the runner's real `GITHUB_OUTPUT` and never removes its `mktemp` repos; the test loop and `harden_security` job are duplicated across the two workflows; `test` and `vulnerability` declare `needs: version` without consuming its outputs.

Codex also traced ref names through the version outputs, xgo linker flags, and the release action and found no command-injection path: values are passed as action inputs, not interpolated into `run` blocks, and the actions are SHA-pinned.

PR #81, opened 2026-09-03
chain,protocol: replace pool copy of a block when momentum bytes differ

0x3639:
The Alphanet stall at momentum 14103620 was triggered by two differently signed copies of the same user transaction (account block). Both signatures were valid. The copies had the same hash and height because the signature is excluded from the block hash, but their stored bytes differed and those bytes contribute to the momentum's changes-hash. The momentum producer used one copy; nodes that received the other first kept it in their pools, computed a different changes-hash, and stopped syncing. This PR makes forced insertion use the momentum's copy and restores affected pool state if insertion fails before commit. Broadcast inserts still keep the first copy. Prefetched blocks are bounded, checked in content order, and copied before application; previously tolerated out-of-order lists are rejected.

The existing [regression test](https://github.com/0x3639/go-zenon/blob/2be586f0051f51bcd99e63461bf38e6b6809713e/protocol/chain_bridge_test.go#L35) reproduces the same byte mismatch by changing the pool copy's unhashed `ChangesHash` field. Verified with Go 1.25.4: it fails against the old insertion logic at `667a69d` with `momentum changes-hash is different than the one computed` and passes at PR head `2be586f`. Run on the PR branch:

```sh
GOWORK=off go test ./protocol -run '^TestInsertChain_MomentumBytesReplacePoolCopyOfSameBlock$' -count=1
```

<details>
<summary>Reproduce the failure before the fix</summary>

From a checkout containing the PR commit, run:

```sh
pr=2be586f0051f51bcd99e63461bf38e6b6809713e
before=$(mktemp -d)
git archive 667a69d9e9a418edf7580b08492ba5dcb9efd63a | tar -x -C "$before"
git show "$pr":protocol/chain_bridge_test.go > "$before/protocol/chain_bridge_test.go"
git show "$pr":chain/nom/account_block.go > "$before/chain/nom/account_block.go"
(cd "$before" && GOWORK=off go test ./protocol -run '^TestInsertChain_MomentumBytesReplacePoolCopyOfSameBlock$' -count=1)
```

The temporary copy includes the unchanged regression test and `AccountBlock.EqualBytes`, its serialization-comparison helper. That helper is the only production-file difference from the base; the account-pool and sync insertion logic remain unchanged. The run should fail with the changes-hash error above. Your checkout is untouched.

</details>

Review limitations remain: readers can observe provisional pool state during validation ([#100](https://github.com/zenon-network/go-zenon/issues/100)); maintainer acceptance is still pending. The existing post-commit test injects its error after broadcast and does not cover the [internal commit-to-broadcast failure window](https://github.com/zenon-network/go-zenon/pull/81#issuecomment-5575192433).

### DIFF
--- a/protocol/chain_bridge.go
+++ b/protocol/chain_bridge.go
@@ -53,6 +53,19 @@ func (c chainBridge) AddAccountBlocks(blocks []*nom.AccountBlock) error {
 	}
 	return nil
 }
+
+// poolHoldsSameBytes reports whether the account pool's frontier chain already
+// contains this exact block. A pool copy that matches by identifier but not by
+// bytes must not be reused, because the stored bytes feed into the momentum
+// changes-hash and the momentum was built from the copy passed in here.
+func (c chainBridge) poolHoldsSameBytes(block *nom.AccountBlock) bool {
+	stored, err := c.chain.GetFrontierAccountStore(block.Address).ByHeight(block.Height)
+	if err != nil || stored == nil {
+		return false
+	}
+	return stored.EqualBytes(block)
+}
+
 func (c chainBridge) GetTransactions() []*nom.AccountBlock {
 	blocks := c.chain.GetAllUncommittedAccountBlocks()
 	return blocks
@@ -184,7 +197,7 @@ func (c chainBridge) InsertChain(momentums []*nom.DetailedMomentum) (int, error)
 			if block.BlockType == nom.BlockTypeContractSend {
 				continue
 			}
-			if patch := c.chain.GetPatch(block.Address, block.Identifier()); patch != nil {
+			if c.poolHoldsSameBytes(block) {
 				// already applied
 				continue
 			}
--- a/protocol/chain_bridge.go
+++ b/protocol/chain_bridge.go
@@ -157,6 +157,25 @@ func (c chainBridge) InsertChain(momentums []*nom.DetailedMomentum) (int, error)
 	}
 	momentums = momentums[start:]
 
+	// Cheap structural check on peer-supplied data before anything is
+	// sized from it, before any rollback and before any pool state is
+	// touched. The verifier repeats the semantic checks later; this only
+	// bounds the input and puts the blocks into content order. The caller
+	// keeps using its own objects concurrently (the fetcher broadcasts the
+	// same DetailedMomentum while importing it), and the VM writes computed
+	// plasma fields onto whatever block it applies, so the loop below works
+	// on private copies of the blocks rather than on the caller's.
+	candidates := make([]*nom.DetailedMomentum, len(momentums))
+	for index, detailed := range momentums {
+		ordered, err := validatePrefetchedBlocks(detailed)
+		if err != nil {
+			log.Error("malformed prefetched account-blocks", "reason", err, "momentum-identifier", detailed.Momentum.Identifier())
+			return index + start, err
+		}
+		candidates[index] = &nom.DetailedMomentum{Momentum: detailed.Momentum, AccountBlocks: copyAccountBlocks(ordered)}
+	}
+	momentums = candidates
+
 	head := momentums[0].Momentum
 	tail := momentums[len(momentums)-1].Momentum
 	ourFrontier, err := store.GetFrontierMomentum()
@@ -213,6 +232,61 @@ func (c chainBridge) InsertChain(momentums []*nom.DetailedMomentum) (int, error)
 	}
 
 	return 0, nil
+}
+
+// validatePrefetchedBlocks rejects a detailed momentum whose block list is not
+// exactly the set of blocks named by the momentum's content: no nil entries,
+// no duplicates, no more entries than a momentum may hold, and one block per
+// content header. Peers are not required to send the blocks in content order,
+// so on success it returns the blocks in content order, which is also the
+// order they must be applied in. The supplied momentum is not modified.
+func validatePrefetchedBlocks(detailed *nom.DetailedMomentum) ([]*nom.AccountBlock, error) {
+	if detailed == nil || detailed.Momentum == nil {
+		return nil, errors.Errorf("missing momentum")
+	}
+	blocks := detailed.AccountBlocks
+	content := detailed.Momentum.Content
+	if len(blocks) > chain.MaxAccountBlocksInMomentum {
+		return nil, errors.Errorf("too many prefetched account-blocks: %v > %v", len(blocks), chain.MaxAccountBlocksInMomentum)
+	}
+	if len(blocks) != len(content) {
+		return nil, errors.Errorf("prefetched account-blocks (%v) do not match momentum content (%v)", len(blocks), len(content))
+	}
+
+	byIdentifier := make(map[types.HashHeight]*nom.AccountBlock, len(blocks))
+	for index, block := range blocks {
+		if block == nil {
+			return nil, errors.Errorf("prefetched account-block at index %v is nil", index)
+		}
+		identifier := block.Identifier()
+		if _, seen := byIdentifier[identifier]; seen {
+			return nil, errors.Errorf("duplicate prefetched account-block %v", identifier)
+		}
+		byIdentifier[identifier] = block
+	}
+
+	ordered := make([]*nom.AccountBlock, len(content))
+	for index, header := range content {
+		if header == nil {
+			return nil, errors.Errorf("momentum content header at index %v is nil", index)
+		}
+		block, ok := byIdentifier[header.Identifier()]
+		if !ok || block.Address != header.Address {
+			return nil, errors.Errorf("momentum content header %v has no matching prefetched account-block", header)
+		}
+		ordered[index] = block
+	}
+	return ordered, nil
+}
+
+// copyAccountBlocks returns deep copies of the blocks, descendants included,
+// so nothing downstream writes to objects the caller still shares.
+func copyAccountBlocks(blocks []*nom.AccountBlock) []*nom.AccountBlock {
+	copied := make([]*nom.AccountBlock, len(blocks))
+	for index, block := range blocks {
+		copied[index] = block.Copy()
+	}
+	return copied
 }
 
 // momentumAddresses lists each account whose pool state the momentum's blocks
--- a/protocol/chain_bridge.go
+++ b/protocol/chain_bridge.go
@@ -2,6 +2,7 @@ package protocol
 
 import (
 	"fmt"
+	"sync"
 
 	"github.com/pkg/errors"
 
@@ -193,34 +194,66 @@ func (c chainBridge) InsertChain(momentums []*nom.DetailedMomentum) (int, error)
 
 	// Insert momentum now
 	for index, detailed := range momentums {
-		for _, block := range detailed.AccountBlocks {
-			if block.BlockType == nom.BlockTypeContractSend {
-				continue
+		// Blocks are force-inserted into the pool before the momentum itself
+		// is validated, so keep a snapshot of every account the momentum
+		// touches and put it back if anything fails.
+		snapshot := c.chain.SnapshotUncommitted(insert, momentumAddresses(detailed))
+		frontierBefore := c.chain.GetFrontierMomentumStore().Identifier()
+		if err := c.insertMomentum(insert, detailed); err != nil {
+			if c.chain.GetFrontierMomentumStore().Identifier() != frontierBefore {
+				// The momentum was committed before the error surfaced. The
+				// snapshot predates that commit, so putting it back would
+				// resurrect pool state that no longer links to the chain.
+				log.Error("momentum committed despite insert error; not restoring account-pool", "reason", err, "momentum-identifier", detailed.Momentum.Identifier())
+			} else if restoreErr := c.chain.RestoreUncommitted(insert, snapshot); restoreErr != nil {
+				log.Error("error while restoring account-pool after failed momentum", "reason", restoreErr, "momentum-identifier", detailed.Momentum.Identifier())
 			}
-			if c.poolHoldsSameBytes(block) {
-				// already applied
-				continue
-			}
-			transaction, err := c.supervisor.ApplyBlock(block)
-			if err != nil {
-				log.Error("error while applying account-block", "reason", err, "account-block-header", block.Header())
-				return index + start, err
-			}
-			if err := c.chain.ForceAddAccountBlockTransaction(insert, transaction); err != nil {
-				log.Error("error while inserting account-block in pool", "reason", err, "account-block-header", block.Header())
-				return index + start, err
-			}
-		}
-
-		transaction, err := c.supervisor.ApplyMomentum(detailed)
-		if err != nil {
-			return index + start, err
-		}
-		if err := c.chain.AddMomentumTransaction(insert, transaction); err != nil {
-			log.Error("error while inserting momentum", "reason", err, "momentum-identifier", detailed.Momentum.Identifier())
 			return index + start, err
 		}
 	}
 
 	return 0, nil
+}
+
+// momentumAddresses lists each account whose pool state the momentum's blocks
+// can modify.
+func momentumAddresses(detailed *nom.DetailedMomentum) []types.Address {
+	addresses := make([]types.Address, 0, len(detailed.AccountBlocks))
+	for _, block := range detailed.AccountBlocks {
+		addresses = append(addresses, block.Address)
+	}
+	return addresses
+}
+
+// insertMomentum applies the momentum's blocks to the pool, then validates and
+// commits the momentum. Under the caller's insert lock.
+func (c chainBridge) insertMomentum(insert sync.Locker, detailed *nom.DetailedMomentum) error {
+	for _, block := range detailed.AccountBlocks {
+		if block.BlockType == nom.BlockTypeContractSend {
+			continue
+		}
+		if c.poolHoldsSameBytes(block) {
+			// already applied
+			continue
+		}
+		transaction, err := c.supervisor.ApplyBlock(block)
+		if err != nil {
+			log.Error("error while applying account-block", "reason", err, "account-block-header", block.Header())
+			return err
+		}
+		if err := c.chain.ForceAddAccountBlockTransaction(insert, transaction); err != nil {
+			log.Error("error while inserting account-block in pool", "reason", err, "account-block-header", block.Header())
+			return err
+		}
+	}
+
+	transaction, err := c.supervisor.ApplyMomentum(detailed)
+	if err != nil {
+		return err
+	}
+	if err := c.chain.AddMomentumTransaction(insert, transaction); err != nil {
+		log.Error("error while inserting momentum", "reason", err, "momentum-identifier", detailed.Momentum.Identifier())
+		return err
+	}
+	return nil
 }
--- a/protocol/chain_bridge.go
+++ b/protocol/chain_bridge.go
@@ -160,21 +160,14 @@ func (c chainBridge) InsertChain(momentums []*nom.DetailedMomentum) (int, error)
 	// Cheap structural check on peer-supplied data before anything is
 	// sized from it, before any rollback and before any pool state is
 	// touched. The verifier repeats the semantic checks later; this only
-	// bounds the input and puts the blocks into content order. The caller
-	// keeps using its own objects concurrently (the fetcher broadcasts the
-	// same DetailedMomentum while importing it), and the VM writes computed
-	// plasma fields onto whatever block it applies, so the loop below works
-	// on private copies of the blocks rather than on the caller's.
-	candidates := make([]*nom.DetailedMomentum, len(momentums))
+	// bounds the input and checks that the blocks are the momentum's
+	// content, in content order.
 	for index, detailed := range momentums {
-		ordered, err := validatePrefetchedBlocks(detailed)
-		if err != nil {
+		if err := validatePrefetchedBlocks(detailed); err != nil {
 			log.Error("malformed prefetched account-blocks", "reason", err, "momentum-identifier", detailed.Momentum.Identifier())
 			return index + start, err
 		}
-		candidates[index] = &nom.DetailedMomentum{Momentum: detailed.Momentum, AccountBlocks: copyAccountBlocks(ordered)}
 	}
-	momentums = candidates
 
 	head := momentums[0].Momentum
 	tail := momentums[len(momentums)-1].Momentum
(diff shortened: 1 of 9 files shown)


georgezgeorgez:
Based on a quick analysis, this hotfix is directionally correct.

However, it replaces the pool’s account block before the momentum has been fully verified. If verification then fails, the replacement and removal of descendant blocks are not reverted.

The correct approach would be to apply replacements to temporary state, verify the resulting momentum and ChangesHash, and commit those changes only if validation succeeds.

It might be more performant to compare only fields excluded from the account-block hash (such as the signature). But full byte comparison is easier to reason about and avoids having to maintain that list as the code evolves.

The bigger architectural issue revealed here is that consensus behavior currently depends on implementation-level database and protobuf serialization. Longer term, the protocol should define a canonical commitment independent of implementation details.

0x3639:
Thanks for the review. Agree on all three points; some detail on the first.

**Replacement before full verification.** This is true, and it is also how the existing path already behaves for every block a momentum carries: `InsertChain` applies each block through the VM and force-adds it into the pool (rolling back any competing block and its descendants at that height) before `ApplyMomentum` runs, and the momentum's changes-hash, signature and producer are only checked in `packMomentum`. Nothing in the protocol layer verifies the momentum earlier. This change extends that existing behaviour to the same-identifier / different-bytes case rather than introducing it.

The exposure is also bounded: the replacement only happens for a copy that passes `verifier.AccountBlock`, so only the holder of the account key can produce a byte-different copy that gets accepted. That party can already rewrite its own pending chain on other nodes through ordinary replacement. And because the momentum's copy now always wins, a pool left holding the "wrong" copy after a failed momentum is corrected by the next momentum that includes the block.

I agree the right fix is to build the momentum against temporary pool state and commit only after `MomentumTransaction` verification passes. That reaches into how the pool and the VM apply path share the memdb managers and is a larger change than this hotfix; happy to take it as a follow-up. A cheaper intermediate step would be to verify momentum hash, signature and producer before the block loop, so only a momentum from the elected pillar can trigger any pool mutation.

**Byte comparison.** Kept it for the reason you give; the cost is ~1.5 µs per pool-resident block and only runs during live sync.

**Canonical commitment.** Fully agree. The changes-hash today is a hash of a protobuf-and-LevelDB-shaped patch, so any stored-but-unhashed field, and any serialization change, is a consensus surface. Defining the commitment over hashed content only is the real fix; this PR just stops the node from getting stuck on the current one.

0x3639:
Correction to my earlier reply: the "only the account key holder can produce an accepted byte-different copy" claim is wrong.

A user block's `ChangesHash` is stored in the serialized block but is neither part of the block hash nor validated anywhere on the apply path: `packBlock` only sets it when signing a locally generated block, and `verifier.AccountBlock` never reads it. So a peer can take a valid signed block, change that one field, and it still passes full block verification. Combined with a genuine signed momentum header, that reaches the forced replacement in this PR, the momentum then fails its changes-hash check, and the pool is left holding the tampered copy with the account's pending descendants dropped. A node that already held the canonical copy would previously have kept it and just rejected the momentum, so as written this PR widens the exposure rather than only closing it. The same field also makes the original stall reproducible without any key, by relaying a tweaked copy to a node ahead of its momentum.

Pre-verifying the momentum header does not help, since the attacker reuses a genuine one.

I'll follow up on this branch with the approach you suggested: snapshot the uncommitted state of every address the momentum touches before the block loop, and restore it on any failure through momentum commit, reusing the pool's existing rebuild path. That also covers the pre-existing case where a different-hash block from a rejected momentum is force-inserted today. Regression tests will assert byte identity and descendant preservation after a failed insertion for both the tampered-`ChangesHash` case and the different-hash case.

0x3639:
Pushed 37d07c7 on top of 00fc755: pool mutations for a momentum are now transactional. Every address the momentum touches has its pool manager cloned before any block is applied, and the clones are installed back on any failure through momentum commit, unless the momentum frontier advanced during the failed insert. Restore is by cloning rather than by re-adding blocks and patches, because re-adding fails for batched embedded transactions and appends frontier writes to stored patches on every replay; the post-momentum rebuild has those two defects today and is tracked in #82. Regression tests cover an altered-ChangesHash copy inside a genuine momentum, a competing pool chain, a second-block failure after a first replacement, a failure after commit, and a batched embedded receive with patch-dump identity. PR description updated to cover both commits.

protocol/chain_bridge.go
```diff
@@ -180,34 +194,66 @@ func (c chainBridge) InsertChain(momentums []*nom.DetailedMomentum) (int, error)
...
-				log.Error("error while applying account-block", "reason", err, "account-block-header", block.Header())
-				return index + start, err
-			}
-			if err := c.chain.ForceAddAccountBlockTransaction(insert, transaction); err != nil {
-				log.Error("error while inserting account-block in pool", "reason", err, "account-block-header", block.Header())
-				return index + start, err
+		// Blocks are force-inserted into the pool before the momentum itself
+		// is validated, so keep a snapshot of every account the momentum
+		// touches and put it back if anything fails.
+		snapshot := c.chain.SnapshotUncommitted(insert, momentumAddresses(detailed))
```
edgepillar, line 220: [P2] Validate and bound prefetched blocks before snapshotting

`momentumAddresses` and `SnapshotUncommitted` allocate based on `len(detailed.AccountBlocks)`, and each existing account manager then clones its `previous`, `versions`, and `patches` maps. This work happens before raw momentum validation limits `Content` to 100 entries and verifies its relationship to the prefetched blocks.

A peer can therefore supply an oversized or duplicate-heavy `AccountBlocks` list within the 10 MiB protocol-message limit and make the node perform additional allocation and cloning while holding the global insert lock. Duplicate entries are especially relevant because the later verifier collapses them into an identifier map, so the raw list itself is not bounded by the `Content` limit.

Please perform a cheap structural preflight before taking the snapshot: reject nil entries, require `len(AccountBlocks) <= MaxAccountBlocksInMomentum`, and enforce an exact one-to-one mapping with `Momentum.Content`. A regression test should also verify that `SnapshotUncommitted` is never called for malformed or oversized input.

protocol/chain_bridge.go
```diff
@@ -180,34 +194,66 @@ func (c chainBridge) InsertChain(momentums []*nom.DetailedMomentum) (int, error)
...
+		frontierBefore := c.chain.GetFrontierMomentumStore().Identifier()
+		if err := c.insertMomentum(insert, detailed); err != nil {
```
edgepillar, line 222: [P2] Keep provisional pool state invisible until validation succeeds

This provides rollback on failure, but it does not provide isolation. `SnapshotUncommitted` releases `ap.changes` before `insertMomentum` force-adds blocks to the live account pool, and `RestoreUncommitted` acquires that lock again only after validation has failed.

During that interval, readers that do not acquire the global insert lock can observe the provisional state. In particular, `GetTransactions` and the unconfirmed/frontier RPC methods can return the replacement block, while new-peer transaction sync can retain that block and send it even after the pool has been restored. Descendants removed by the provisional replacement are similarly hidden during the window.

Please apply the candidate mutations to detached state and publish them atomically only after momentum validation succeeds, or otherwise prevent all pool readers from observing provisional managers. A deterministic regression test could pause immediately after the force-add and assert that concurrent transaction-sync and RPC-style readers still see the original pool state.

edgepillar (commented):
The byte-level comparison and forced-only replacement appear to address the original liveness failure, and the follow-up commit prevents persistent pool corruption on the ordinary pre-commit failure paths. Before merging, I think the two inline concerns above should either be addressed or explicitly accepted by the maintainer. I would also add coverage for a failure after `chainManager.Add` but before event broadcast, plus a direct clone-divergence test for `memdbManager.Clone`.

0x3639:
Thanks for the careful read. Both inline points are accurate; here is what I did with each.

**Bound the prefetched blocks before snapshotting.** Agreed and done in the new commit. As soon as the already-known momentums are trimmed, and before the side-chain rollback decision or any pool work, `InsertChain` now rejects a detailed momentum unless its block list has no nil entries, no duplicates, at most `MaxAccountBlocksInMomentum` entries, and exactly one block per content header matched by identifier and address. Order is deliberately not required, since the verifier accepts any order through its identifier map and a positional rule would reject consensus-valid momentums and cost the peer its connection; on success the blocks are returned in content order, which is the order they must be applied in anyway, and `InsertChain` works from a private copy of each momentum with deep-copied blocks rather than touching the caller's objects, since the fetcher broadcasts the same `DetailedMomentum` concurrently with the import and the VM writes computed plasma fields onto the block it applies. That VM write is pre-existing for unknown blocks, but the byte-aware replacement in this PR sends pool-resident blocks through the VM too, so the copy closes a race this branch would otherwise have widened. The verifier still does the semantic checks afterwards. One note on scope: the pre-existing loop already ran full VM apply and signature verification on every entry of that unbounded list before the content check, so the preflight protects the old code at least as much as the snapshot. Tests: `TestInsertChain_RejectsMalformedPrefetchedBlocksBeforeTouchingPool` and `TestInsertChain_RejectsDuplicatePrefetchedBlocks` feed the malformed shapes through a chain wrapper that counts snapshots and assert no snapshot, no pool change and no frontier change; `TestInsertChain_AcceptsPrefetchedBlocksInAnyOrder` inserts a two-account momentum with the list swapped; `TestInsertChain_MalformedSideChainDoesNotRollBack` offers a longer side chain whose first momentum is oversized and checks the frontier is untouched.

**Keep provisional state invisible until validation succeeds.** Also accurate: between the force-add and the restore, readers that take only the pool lock (the ledger RPC methods, the pillar workers, `GetTransactions` for peer sync) can observe the provisional state for the duration of one insert. I'd ask the maintainer to accept this for the hotfix rather than fix it here, for two reasons. It is not new: the pre-existing path force-added every unknown block before momentum validation with exactly the same visibility. And the fix is the temporary-state design georgezgeorgez described, where candidate mutations go to detached managers and the VM apply path, the momentum apply and every reader agree on which set they see; that reaches into how the pool is read throughout the node and is the follow-up I've already committed to. Under this PR the residual effect is transient and self-correcting: a peer that picks up a provisional copy during the window has it replaced when the real momentum arrives, and a failed momentum leaves the local pool byte-identical to before.

**Extra tests.** A failure after `chainManager.Add` but before the event broadcast is what `TestInsertChain_FailureAfterCommitDoesNotRestoreStalePool` exercises: it wraps the chain so `AddMomentumTransaction` commits and then errors, which from the bridge's side is indistinguishable from the internal prefetch failure, and asserts the stale snapshot is not put back. If you'd rather see the internal variant I can add it. The direct clone-divergence test is added as `TestMemDBManagerCloneDivergesIndependently`: Add and Pop on either side of the clone are invisible to the other, and the shared patch's bytes are unchanged.

protocol/chain_bridge.go
0x3639: Addressed in 0884a3f. `InsertChain` now validates every remaining momentum right after the already-known ones are trimmed, before the side-chain rollback decision and before any snapshot or pool work: no nil entries, no duplicates, at most `MaxAccountBlocksInMomentum` entries, and exactly one block per content header matched by identifier and address. Order is deliberately not required, since the verifier accepts any order through its identifier map and a positional rule would reject consensus-valid momentums and cost the peer its connection; the blocks are returned in content order, which is the order they are applied in. The loop then runs on private copies with deep-copied blocks, because the fetcher broadcasts the same `DetailedMomentum` concurrently with the import and the VM writes plasma fields onto the block it applies. Tests: `TestInsertChain_RejectsMalformedPrefetchedBlocksBeforeTouchingPool`, `TestInsertChain_RejectsDuplicatePrefetchedBlocks`, `TestInsertChain_AcceptsPrefetchedBlocksInAnyOrder`, `TestInsertChain_MalformedSideChainDoesNotRollBack`, `TestInsertChain_DoesNotMutateCallerDetailedMomentum`, `TestInsertChain_DoesNotWriteToCallerAccountBlocks`. The clone-divergence test you asked for is `TestMemDBManagerCloneDivergesIndependently`.

protocol/chain_bridge.go
0x3639: Accurate, and not addressed in this PR. Between the force-add and the restore, readers that take only the pool lock (ledger RPC, pillar workers, `GetTransactions` for peer sync) can observe the provisional state for the duration of one insert. Two things worth weighing: this is the same visibility the existing path already had, since it force-added every unknown block before momentum validation; and under this PR the effect is transient and self-correcting, because a peer that picks up a provisional copy has it replaced when the real momentum arrives, and a failed momentum leaves the local pool byte-identical to before. Real isolation is the detached-state design georgezgeorgez described, where candidate mutations go to staging managers and the VM apply path, the momentum apply and every reader agree on which set they see. That reaches into how the pool is read throughout the node, so I would ask the maintainer to accept this for the hotfix and take it as the follow-up, which I have committed to. On the post-commit test: `TestInsertChain_FailureAfterCommitDoesNotRestoreStalePool` wraps the chain so `AddMomentumTransaction` commits and then errors, which is what the bridge sees for the internal prefetch failure too; happy to add the internal variant if you would rather have it explicitly.

digitalSloth:
@0x3639 Nice work, looks like its almost there, these were the only things that came back from my review

**Major — please address:**

1. `chain_bridge_test.go:20-142` — the hand-rolled Ed25519/`math/big` signature forgery (~140 lines) isn't needed. `ChangesHash` is serialized but excluded from `ComputeHash` and never validated for user blocks (`vm/vm.go:106` only covers `BlockTypeContractReceive`; the verifier has no account-block check), so a mutated `ChangesHash` alone gives the same identifier, still verifies, still enters the pool, and reproduces the byte-different-copy bug without a private key. Confirmed by reverting the fix and running both fixtures: a `ChangesHash`-mutated copy fails with the same `momentum changes-hash is different than the one computed` as the signature-forged one. Recommend dropping the crypto code (`chain_bridge_test.go:20-142`, and the matching helper in `chain_bridge_restore_test.go:414`) in favor of a 5-line `ChangesHash` swap.

2. The batch-wide deep copy at `chain_bridge.go:168-177` holds up to ~25,600 `AccountBlock.Copy()` results live per `InsertChain` call — `maxBlockProcess` is 256 and the preflight caps each momentum at 100 blocks. Keep the preflight validation up front, but move `copyAccountBlocks` into the per-momentum insert loop, so a batch that fails on momentum 3 stops paying for copies on the other 253.

**Minor — worth a look, not blocking:**

3. `validatePrefetchedBlocks` (`chain_bridge.go:256-278`) reorders peer input into content order. Every producer in the repo already emits content order (`PrefetchMomentum`, `chainBridge.GetBlock` both index by `momentum.Content`, and RLP preserves it), so an index-wise `blocks[i]` vs `content[i]` check would be allocation-free and stricter. That said, the reorder isn't unsafe — the list is still exactly the momentum's content with no nils or duplicates — it's tolerance for a peer implementation outside this repo, at the cost of one ≤100-entry map. Open to switching if you'd rather have the stricter contract.

0x3639:
All three addressed in 2be586f.

1. Confirmed independently before removing the crypto: a `ChangesHash`-only swap on the pool copy fails with `momentum changes-hash is different than the one computed` with the pool-side byte check disabled, and passes with it. The Ed25519 code and the helper in `chain_bridge_restore_test.go` are gone; both tests use a small `byteDifferentCopy` helper that asserts the identifier is kept and the bytes differ.

2. `copyAccountBlocks` now runs inside the insert loop, one momentum at a time. The structural preflight still covers the whole batch before any rollback or pool work.

3. Switched to the strict index-wise check. One thing the map was also doing: rejecting a content header listed twice before anything touched the pool. The verifier catches that shape too, but only after a force-add and snapshot restore, so I kept that guarantee with an allocation-free pairwise scan over the content (≤100 entries). `TestInsertChain_AcceptsPrefetchedBlocksInAnyOrder` became `TestInsertChain_RejectsPrefetchedBlocksOutOfContentOrder`, and the duplicate test gained a duplicated-content case that fails without the scan.

edgepillar:
Thanks for the follow-up in `2be586f`. The per-momentum copy now limits eager copying while retaining the batch-wide structural preflight. The smaller byte-mismatch fixture also avoids unnecessary cryptographic test machinery.

Two points remain worth recording before merge:

1. The [strict positional check](https://github.com/zenon-network/go-zenon/blob/2be586f0051f51bcd99e63461bf38e6b6809713e/protocol/chain_bridge.go#L247-L277) intentionally narrows the previously tolerated prefetched-block layouts. The PR description still says order is not required and lists the old any-order test and signature-generation fixture. Could it be updated to match the current implementation and explicitly document this sync-input compatibility decision?
2. As acknowledged in the [isolation discussion](https://github.com/zenon-network/go-zenon/pull/81#discussion_r3933747924), rollback repairs local state after failure but does not prevent RPC, pillar, or peer-sync readers from observing provisional state beforehand. Could the maintainer explicitly record whether that residual visibility is acceptable for this hotfix and link the detached-state/atomic-publication follow-up?

I am not asking to fold the larger redesign into this PR or claiming this visibility is newly introduced. The important distinction is failure restoration versus isolation; the current change should not be described as providing both.

This remains a source-only review. I have not executed the tests or independently verified the reported validation.

0x3639:
Both points are recorded.

1. PR description updated. Commit 3 now describes only the bound and the one-to-one mapping; a new commit 4 section covers 2be586f, including the content-order decision as an explicit narrowing of accepted sync input, why it reverses my earlier position in this thread (every producer in the repository emits content order, so the tolerance covered a peer implementation that does not exist), and what a rejection costs the peer on each path (downloader: peer dropped and sync round cancelled; fetcher: logged and ignored). The tests section lists the current test names; the any-order test and the signature fixture are gone from it.

2. The description now has a section stating that the snapshot is failure restoration, not isolation, with the visibility window and its readers spelled out. The detached-state / atomic-publication work is opened as #100 with the thread linked and the acceptance test you described. Whether the residual visibility is acceptable for the hotfix is left as the maintainer's call, and the description says so.

edgepillar:
One additional source-level gap is worth covering before merge.

**[P2] Exercise the actual commit-to-broadcast failure window**

[`TestInsertChain_FailureAfterCommitDoesNotRestoreStalePool`](https://github.com/zenon-network/go-zenon/blob/2be586f0051f51bcd99e63461bf38e6b6809713e/protocol/chain_bridge_restore_test.go#L183-L223) calls the real `AddMomentumTransaction` to completion—including `PrefetchMomentum` and `broadcastInsertMomentum`—and only then returns `errAfterCommit`. The internal partial-commit path has a materially different publication state: [`momentumPool.AddMomentumTransaction`](https://github.com/zenon-network/go-zenon/blob/2be586f0051f51bcd99e63461bf38e6b6809713e/chain/momentum_pool.go#L28-L50) advances `chainManager` first, can return from `PrefetchMomentum` with an error, and has not yet broadcast the momentum to `accountPool` or the other listeners.

The bridge is right not to restore a snapshot once the frontier advanced, because that snapshot is stale. The current test establishes that decision, but it does not establish that the pool and listeners are coherent when the real pre-broadcast failure occurs. In that case the stable momentum frontier has advanced while the account-pool rebase and all event delivery were skipped. This path predates the PR; the concern is that the current regression test is not equivalent to it.

Could you either add deterministic fault injection at that exact boundary or track the post-commit publication/recovery contract explicitly? I would assert the committed frontier, the pool's stable/uncommitted view, and listener delivery state, rather than only the committed block bytes. This can remain separate from the provisional-read isolation work in #100.

This is a source-only finding; I have not run the tests.

digitalSloth (approved):

0x3639:
The Alphanet stall at momentum 14103620 was triggered by two differently signed copies of the same user transaction (account block). Both signatures were valid. The copies had the same hash and height because the signature is excluded from the block hash, but their stored bytes differed and those bytes contribute to the momentum's changes-hash. The momentum producer used one copy; nodes that received the other first kept it in their pools, computed a different changes-hash, and stopped syncing. This PR makes forced insertion use the momentum's copy and restores affected pool state if insertion fails before commit. Broadcast inserts still keep the first copy. Prefetched blocks are bounded, checked in content order, and copied before application; previously tolerated out-of-order lists are rejected.

The existing [regression test](https://github.com/0x3639/go-zenon/blob/2be586f0051f51bcd99e63461bf38e6b6809713e/protocol/chain_bridge_test.go#L35) reproduces the same byte mismatch by changing the pool copy's unhashed `ChangesHash` field. Verified with Go 1.25.4: it fails against the old insertion logic at `667a69d` with `momentum changes-hash is different than the one computed` and passes at PR head `2be586f`. Run on the PR branch:

```sh
GOWORK=off go test ./protocol -run '^TestInsertChain_MomentumBytesReplacePoolCopyOfSameBlock$' -count=1
```

<details>
<summary>Reproduce the failure before the fix</summary>

From a checkout containing the PR commit, run:

```sh
pr=2be586f0051f51bcd99e63461bf38e6b6809713e
before=$(mktemp -d)
git archive 667a69d9e9a418edf7580b08492ba5dcb9efd63a | tar -x -C "$before"
git show "$pr":protocol/chain_bridge_test.go > "$before/protocol/chain_bridge_test.go"
git show "$pr":chain/nom/account_block.go > "$before/chain/nom/account_block.go"
(cd "$before" && GOWORK=off go test ./protocol -run '^TestInsertChain_MomentumBytesReplacePoolCopyOfSameBlock$' -count=1)
```

The temporary copy includes the unchanged regression test and `AccountBlock.EqualBytes`, its serialization-comparison helper. That helper is the only production-file difference from the base; the account-pool and sync insertion logic remain unchanged. The run should fail with the changes-hash error above. Your checkout is untouched.

</details>

Review limitations remain: readers can observe provisional pool state during validation ([#100](https://github.com/zenon-network/go-zenon/issues/100)); maintainer acceptance is still pending. The existing post-commit test injects its error after broadcast and does not cover the [internal commit-to-broadcast failure window](https://github.com/zenon-network/go-zenon/pull/81#issuecomment-5575192433).

georgezgeorgez (approved):

openmac0815:
**Review — APPROVE (verified), consensus-critical**

This addresses the momentum-14103620 stall class: two validly-signed copies of one account block, same hash/height (signature excluded from hash), different stored bytes feeding the momentum changes-hash. Verified in the diff:

- `EqualBytes` comparison at the pool's duplicate check; forced inserts replace byte-different copies, broadcast inserts still keep first copy.
- `validatePrefetchedBlocks` bounds and order-checks peer-supplied block lists before anything is sized from them — nil entries, count mismatch, duplicates all rejected pre-rollback.
- Per-momentum `SnapshotUncommitted`/`RestoreUncommitted` with `memdbManager.Clone()`; the clone-divergence test pins the shared-base semantics. Notably, restore is correctly skipped when the momentum committed despite the error (the snapshot predates the commit and would resurrect unlinked pool state).
- Private copies of blocks for application — the VM writes computed plasma fields onto applied blocks while the fetcher broadcasts the same objects concurrently.

Merge note: conflicts with #83 (both rewrite `InsertChain` from different bases). Recommend landing this one first.

PR #83, opened 2026-09-04
chain,protocol: keep the canonical branch when a side-chain replacement fails

0x3639:
## Summary

`InsertChain` accepted a side chain on structural grounds alone: the head links to a canonical momentum at most 30 back, and the tail is higher than our frontier. It then rolled the canonical chain and cache back to the fork point before applying the first replacement momentum, and only afterwards ran the checks that need chain state (changes-hash, producer, VM application). A candidate that failed any of those left the node with its canonical branch removed and nothing in its place, until it resynchronised from another peer. Any connected peer selected for a download could trigger this with a momentum that is internally consistent but signed by a key that is not the elected producer, or that carries a changes-hash the node cannot reproduce.

## Changes

Side-chain replacement now leaves the chain in one of three states: fully replaced, on a committed replacement prefix that is longer than the branch it removed, or exactly as it was.

- **Static checks before any rollback.** Every candidate's advertised hash must match its content and its signature must verify under the carried public key. The head is additionally run through the raw momentum verifier against the store at the fork point. A candidate that fails these never causes a rollback. The state-dependent checks still run afterwards, as before.
- **Undo log before the rollback.** `momentumPool.CaptureBranchAbove` records each momentum above the fork point with its prefetched account blocks and the state patch the momentum store holds for it. The ldb-backed manager deletes a patch when its momentum is popped, so the capture happens first; if any patch is missing the insert is refused without touching the chain. The stored patch already carries the three frontier writes `Manager.Add` appends on every insert, so the capture strips them (`db.FrontierWriteKeys`, `db.RemoveKeys`); otherwise each restore would persistently grow the forward and rollback records of up to 30 momentums.
- **Restore on a later failure, while the prefix is short.** If applying the replacement fails at some candidate and the committed prefix is no longer than the removed branch, the bridge rolls the applied candidates back to the fork point and re-adds the captured momentums in order, updating the cache from the stored patch and committing through `AddMomentumTransaction`, so listeners see the same insert events a normal insert produces. A restore failure is handled like the existing uncertain-state cases: logged at Crit, process exits.
- **Keep the prefix once it is longer than what was removed.** The removed branch is bounded to 30 momentums by the depth check, but the committed replacement prefix is bounded only by the downloader batch (up to 256). The cache keeps 100 rollback steps, so a restore after more than 100 committed candidates would roll the canonical chain back and then fail on the cache, exiting with a state `Init` cannot reconcile on restart. Once the prefix is longer than the removed branch, the node is already on a valid chain longer than the one it had, so the prefix stays, which is what the code did before this change. This bounds every restore to at most 30 steps.

What the restore does not cover, deliberately: the rollback still clears the account pool as any rollback does (valid blocks force-inserted from the rejected candidates remain, as if a peer had relayed them); event listeners see delete notifications followed by insert notifications for restored momentums; and the captured branch is held in memory only, so a crash mid-restore leaves the node at the fork point plus whatever was committed, a valid prefix it resynchronises from, not the restored branch. None of this is crash-atomic and it is not meant to be.

## Tests

All in `protocol/chain_bridge_sidechain_test.go`, through the real `InsertChain` on the mock node. Genuine forks are produced on a second mock node sharing genesis and clock, so a fork longer than the cache window can be built without rolling anything back.

- `SideChainRestoredWhenCandidateFailsAfterRollback`: head passes every stateless check, fails after the rollback; frontier, every momentum hash, cache frontier and account balances unchanged.
- `SideChainWithInvalidSignatureIsRejectedBeforeRollback`, `SideChainWithTamperedHashIsRejectedBeforeRollback`: rejected with no rollback at all.
- `ValidLongerSideChainReplacesBranch`: a genuine fork produced by the mock pillars still replaces the branch.
- `SideChainRestoredWhenLaterCandidateFails`: the fork's first momentum is applied, the second candidate fails, the original branch is restored, and the genuine fork then inserts on top of the restored state.
- `RestoreBoundaryAtRemovedBranchLength`: with a two-momentum original branch, a committed prefix of one or two is restored and a prefix of three is kept; in every case the rest of the genuine fork then inserts on top of whatever the node was left with.
- `ReplacementPrefixBeyondCacheWindowIsKept`: 101 replacement momentums commit, the 102nd fails; the prefix is kept with chain and cache frontiers agreeing, and the remaining fork inserts on top. Without the guard this test exits the process with the cache rollback error.
- `SideChainDepthBoundary`: depth 31 rejected untouched; depth 30 rolls back, fails, and all 30 momentums come back.
- `RepeatedRestoresKeepStoredPatchesIdentical`: three consecutive failed attempts leave every stored patch above the fork point byte-identical. The comparison reads the raw stored patches through the new `GetMomentumPatch` accessor, frontier writes included, rather than through `CaptureBranchAbove`, which strips them and would hide the growth.

Verification: `GOWORK=off go build ./...`, `go test -race ./chain ./common/db ./protocol`, `go test ./...`, `golangci-lint run` clean on the touched files.

Ref: internal security review finding on unverified side-chain rollback (affects every tagged release through v0.1.0).

### DIFF
--- a/chain/interface.go
+++ b/chain/interface.go
@@ -37,6 +37,10 @@ type MomentumEventManager interface {
 }
 
 type MomentumPool interface {
+	// CaptureBranchAbove returns every momentum above identifier, lowest first,
+	// together with its stored state patch, so the branch can be re-inserted
+	// after a rollback. It fails, without side effects, if any patch is missing.
+	CaptureBranchAbove(insertLocker sync.Locker, identifier types.HashHeight) ([]*RemovedMomentum, error)
 	AddMomentumTransaction(insertLocker sync.Locker, transaction *nom.MomentumTransaction) error
 	RollbackTo(insertLocker sync.Locker, identifier types.HashHeight) error
 
--- a/chain/interface.go
+++ b/chain/interface.go
@@ -41,6 +41,9 @@ type MomentumPool interface {
 	// together with its stored state patch, so the branch can be re-inserted
 	// after a rollback. It fails, without side effects, if any patch is missing.
 	CaptureBranchAbove(insertLocker sync.Locker, identifier types.HashHeight) ([]*RemovedMomentum, error)
+	// GetMomentumPatch returns the state patch stored for the momentum with
+	// this identifier, as the manager holds it, or nil if there is none.
+	GetMomentumPatch(identifier types.HashHeight) db.Patch
 	AddMomentumTransaction(insertLocker sync.Locker, transaction *nom.MomentumTransaction) error
 	RollbackTo(insertLocker sync.Locker, identifier types.HashHeight) error
 
--- a/chain/momentum_pool.go
+++ b/chain/momentum_pool.go
@@ -166,6 +166,12 @@ func (c *momentumPool) CaptureBranchAbove(insertLocker sync.Locker, identifier t
 	return removed, nil
 }
 
+func (c *momentumPool) GetMomentumPatch(identifier types.HashHeight) db.Patch {
+	c.changes.Lock()
+	defer c.changes.Unlock()
+	return c.chainManager.GetPatch(identifier)
+}
+
 func (c *momentumPool) RollbackTo(insertLocker sync.Locker, identifier types.HashHeight) error {
 	c.log.Info("rollbacking momentums", "to-identifier", identifier)
 	if insertLocker == nil {
--- a/chain/momentum_pool.go
+++ b/chain/momentum_pool.go
@@ -108,6 +108,64 @@ func (c *momentumPool) AddMomentumTransaction(insertLocker sync.Locker, transact
 
 	return nil
 }
+
+// RemovedMomentum is one momentum taken off the canonical chain by a rollback,
+// with the state patch that re-inserts it.
+type RemovedMomentum struct {
+	Detailed *nom.DetailedMomentum
+	Changes  db.Patch
+}
+
+func (c *momentumPool) CaptureBranchAbove(insertLocker sync.Locker, identifier types.HashHeight) ([]*RemovedMomentum, error) {
+	if insertLocker == nil {
+		return nil, errors.Errorf("insertLocker can't be nil")
+	}
+	c.changes.Lock()
+	defer c.changes.Unlock()
+
+	store := c.getFrontierStore()
+	target, err := store.GetMomentumByHeight(identifier.Height)
+	if err != nil {
+		return nil, err
+	}
+	if target == nil || target.Hash != identifier.Hash {
+		return nil, errors.Errorf("can't capture branch above %v. Not on the canonical chain", identifier)
+	}
+	frontier, err := store.GetFrontierMomentum()
+	if err != nil {
+		return nil, err
+	}
+
+	removed := make([]*RemovedMomentum, 0, frontier.Height-identifier.Height)
+	for height := identifier.Height + 1; height <= frontier.Height; height++ {
+		momentum, err := store.GetMomentumByHeight(height)
+		if err != nil {
+			return nil, err
+		}
+		if momentum == nil {
+			return nil, errors.Errorf("can't capture branch above %v. Missing momentum at height %v", identifier, height)
+		}
+		detailed, err := store.PrefetchMomentum(momentum)
+		if err != nil {
+			return nil, err
+		}
+		stored := c.chainManager.GetPatch(momentum.Identifier())
+		if stored == nil {
+			return nil, errors.Errorf("can't capture branch above %v. Missing patch for momentum %v", identifier, momentum.Identifier())
+		}
+		// The stored patch already carries the frontier writes Manager.Add
+		// appended when the momentum was first inserted. Strip them so that
+		// re-adding the momentum appends them exactly once and the stored
+		// patch stays byte-identical across a restore.
+		changes, err := db.RemoveKeys(stored, db.FrontierWriteKeys(momentum.Identifier()))
+		if err != nil {
+			return nil, err
+		}
+		removed = append(removed, &RemovedMomentum{Detailed: detailed, Changes: changes})
+	}
+	return removed, nil
+}
+
 func (c *momentumPool) RollbackTo(insertLocker sync.Locker, identifier types.HashHeight) error {
 	c.log.Info("rollbacking momentums", "to-identifier", identifier)
 	if insertLocker == nil {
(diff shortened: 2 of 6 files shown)


edgepillar:
Thanks for adding recovery when a side-chain replacement fails. I reviewed `064b936` and found a restoration-window mismatch that I think should be addressed before merge.

This is a source-level review; I have not executed a runtime reproducer or independently reproduced the reported test results.

**1. Restoration can exceed the cache rollback window**

The 30-momentum limit bounds the original canonical branch being removed, but it does not bound the successfully committed replacement prefix.

If more than 100 replacement momentums commit and a later candidate fails, `restoreBranch` calls `rollbackSideChain`. That function rolls the canonical database back to the fork point **before** attempting cache rollback. The cache retains only 100 rollback steps, so the latter operation fails after canonical state has already moved. The restoration error then terminates the process.

The relevant paths are [rollback ordering](https://github.com/zenon-network/go-zenon/blob/064b9365888da4711bf5c5be48e08f4bef5c7708/protocol/chain_bridge.go#L27-L37), [restoration and process termination](https://github.com/zenon-network/go-zenon/blob/064b9365888da4711bf5c5be48e08f4bef5c7708/protocol/chain_bridge.go#L234-L239), and [cache rollback-history pruning](https://github.com/zenon-network/go-zenon/blob/064b9365888da4711bf5c5be48e08f4bef5c7708/chain/cache/storage/db.go#L88-L115).

Normal restart also appears unable to reconcile this state: when the cache is ahead of the canonical chain, initialization invokes the same bounded rollback path. See [cache rollback and initialization](https://github.com/zenon-network/go-zenon/blob/064b9365888da4711bf5c5be48e08f4bef5c7708/chain/cache.go#L91-L135).

This is not an unconditional arbitrary-peer crash claim. A network trigger requires a state-valid alternative prefix; valid signatures alone cannot manufacture that prefix. However, a legitimate long replacement followed by a late validation failure exposes the same correctness problem.

Could the replacement operation preserve the rollback information it needs throughout the entire accepted transaction, or validate/apply against detached state? Simply increasing the cache constant would move the boundary without establishing the required invariant.

**2. Regression coverage should cross the replacement-prefix boundary**

Please consider cases with 99, 100, and 101 successfully committed replacement momentums followed by a failing candidate. They should verify restoration of both frontiers and successful reopening of persistent chain/cache state.

The current later-candidate test commits only one replacement before failing, while the depth-30 test concerns the original branch depth. Neither exercises this cache boundary.

It would also help to compare raw stored patches, including frontier metadata, across repeated restores. The current comparison obtains patches through `CaptureBranchAbove`, which strips those metadata writes. Fault injection around cache updates and canonical writes would strengthen the recovery guarantees further.

**3. Clarify the recovery guarantee**

As the PR description acknowledges, restoring the canonical branch is not equivalent to restoring the pending account pool or undoing externally observed events. The captured recovery data is also held in memory, so this should not be described as crash-atomic or isolated publication.

Could those boundaries remain explicit, with event-sequence and pending-pool expectations covered where practical?

My main concern is item 1; the remaining points clarify and test the intended recovery boundary.

0x3639:
Reviewed and reproduced. `TestInsertChain_ReplacementPrefixBeyondCacheWindowIsKept` commits 101 replacement momentums and fails the 102nd; on `064b936` it exits the test binary with "canonical chain rolled back to {…5} but cache rollback failed … outside the rollback cache", which is the sequence you described.

Fix in `fc02653`: `InsertChain` restores only while the committed prefix is no longer than the removed branch. Once the prefix is longer, the node is already on a valid chain longer than the one it had, so the prefix stays, which is what the code did before the restore existed. The depth check bounds the removed branch to 30, so every restore is at most 30 steps and always inside the cache window. This uses the existing bound rather than moving the cache constant.

On the tests: the fork fixture now builds the genuine fork on a second mock node so forks longer than the cache window can be produced, `RestoreBoundaryAtRemovedBranchLength` pins the decision at prefix lengths one below, equal to, and one above the removed branch, and `RepeatedRestoresKeepStoredPatchesIdentical` now compares raw stored patches through a new `GetMomentumPatch` accessor, frontier writes included. With the strip in `CaptureBranchAbove` disabled, the raw comparison reports the patch doubling that the old comparison hid.

On the guarantee: the PR description and the `restoreBranch` comment now say what the restore covers and what it does not: the account pool stays as the rollback left it, listeners see delete then insert events, and the captured branch is in memory only, so a crash mid-restore leaves the node at the fork point plus the committed prefix, not the restored branch. Reopen-after-crash is not tested; the kept-prefix and restored states are checked by inserting the rest of the genuine fork on top of them.

edgepillar:
Thanks for the follow-up. I re-reviewed `fc0265339edd0120aa7ca64eeafa1a1c6b15f9de` against `064b9365888da4711bf5c5be48e08f4bef5c7708`.

The `index > len(removed)` guard addresses the cache-window failure I reported at source level: the removed branch is bounded to 30 momentums, so restoration now rolls back at most 30 committed replacements; a longer valid replacement prefix is retained.

The new below/equal/above boundary cases and the 101-committed/102nd-failing case target that behavior directly. Reading raw stored patches also removes the previous test oracle's filtering blind spot. I found no new blocking issue in this follow-up's keep/restore decision.

This was a static review: I did not run the tests and could not independently verify successful CI. The documented account-pool/event limitations and lack of reopen-after-crash testing remain outside what this review validates.

As non-blocking test hardening, I would still suggest a removed depth of 30 with 30/31 committed replacements, an already-known input prefix (`start > 0`), and a compensated insertion failure after the cache update.

I consider my original cache-window finding addressed in the reviewed source, not a blanket approval of crash recovery or isolation.

openmac0815:
**Review — APPROVE (verified), consensus-critical**

The failure mode is real and serious: `InsertChain` rolled the canonical chain back to the fork point before running state-dependent checks, so a structurally-plausible candidate that then failed (bad changes-hash, wrong producer) left the node with no canonical branch. Verified:

- Static checks (advertised hash vs content, signature vs carried public key) run before any rollback; head additionally verified against the store at the fork point.
- `CaptureBranchAbove` records momentums + patches before rollback; missing patch → insert refused without touching the chain. Frontier-write stripping via `RemoveKeys`/`FrontierWriteKeys` keeps re-added patches byte-identical.
- The three-state outcome (full replace / longer committed prefix / exact restore) is sound; the prefix-stays rule once it exceeds the removed branch is correctly reasoned against the 30-deep fork bound and cache rollback window.
- `log.Crit` + exit on failed restore is drastic but matches the file's existing convention for unrecoverable chain state.

Honest limitations documented in the code: restore doesn't undo emitted delete events and leaves the account pool as rollback left it; the captured branch is memory-only.

Merge note: conflicts with #81 — land #81 first, then rebase this.

PR #84, opened 2026-09-05
protocol: bound the number of hashes one GetBlocksMsg may name

0x3639:
## Summary

The `GetBlocksMsg` handler decoded hashes until the RLP list ended and looked each one up in the momentum store before considering any limit. The only limit inside the loop, `downloader.MaxBlockFetch` (128), was compared against the number of blocks *found*, so a request whose hashes are not in the store never tripped it and ran to the 10 MiB message ceiling, which is more than 300,000 lookups from one message. When nothing was found, the handler then built a debug string over the whole list with repeated string concatenation, which copies O(n²) bytes and runs before the logger checks whether debug output is enabled. Any admitted peer can send this message.

## Changes

- **`protocol.MaxBlocksRequest`** (new, = `fetcher.HashLimit`, 256): the most hashes one `GetBlocksMsg` may name. The handler checks it after each hash decodes and before the hash is appended or looked up. The first hash past the limit ends the handler with `ErrMsgTooLarge`, so the peer is dropped the same way an oversized frame is. The existing reply cap of `MaxBlockFetch` found blocks is unchanged.
- **`fetcher.HashLimit`**: the fetcher's per-peer announce limit is exported (rename only) so the protocol package can reference it. It is the larger of the two honest batch sizes: the downloader requests at most 128 blocks, the fetcher batches up to 256 announced hashes per peer.
- **`peer.RequestBlocks`**, the only place a `GetBlocksMsg` is encoded, splits a batch larger than `MaxBlocksRequest` into consecutive requests of at most that size. The fetcher's announce counter is not exact (`forgetHash` decrements when a hash moves to `fetching` and again on completion), so this makes the bound hold on the sending side regardless. Batches of 256 or fewer are sent exactly as before.
- The all-miss debug line logs the request count and the first hash prefix instead of formatting the whole list.

## Compatibility

Peers on earlier releases do not split, so a request of more than 256 hashes from such a peer is dropped unless the responder finds 128 blocks among the first 256 hashes, in which case the reply cap ends the loop first. A drop therefore needs a miss-heavy oversized request; an announcer holds the blocks it announced. The fetcher counter drift is left for a separate change.

## Tests

`protocol/handler_getblocks_test.go` drives the real `handleMsg` over a `p2p.MsgPipe` with a chain stub that serves fixed blocks and counts `GetBlock` calls:

- 257 and 2048 unknown hashes return `ErrMsgTooLarge`, write no reply, and perform exactly 256 lookups.
- Known hashes interleaved with unknown ones count the same and are rejected at the limit.
- Exactly 256 unknown hashes are legitimate and get an empty reply after 256 lookups.
- 260 known hashes still return 128 blocks after 128 lookups.
- `RequestBlocks` with 1, 256, 257 and 517 hashes produces frames of 1..256 hashes each, complete and in order.
- The constant is pinned at or above both `downloader.MaxBlockFetch` and `fetcher.HashLimit`.

The rejection and splitting tests fail on the unpatched code.

## Verification

`GOWORK=off`: gofmt and vet clean; `golangci-lint run ./protocol/...` adds no issues on changed lines; `go test -race -count=3 ./protocol/...`; `go test ./...` (full suite).

### DIFF
--- a/protocol/fetcher/fetcher.go
+++ b/protocol/fetcher/fetcher.go
@@ -37,7 +37,7 @@ const (
 	fetchTimeout  = 5 * time.Second        // Maximum alloted time to return an explicitly requested block
 	maxUncleDist  = 7                      // Maximum allowed backward distance from the chain head
 	maxQueueDist  = 32                     // Maximum allowed distance from the chain head to queue
-	hashLimit     = 256                    // Maximum number of unique blocks a peer may have announced
+	HashLimit     = 256                    // Maximum number of unique blocks a peer may have announced; also the largest batch one fetch request names
 	blockLimit    = 64                     // Maximum number of unique blocks a per may have delivered
 )
 
@@ -267,8 +267,8 @@ func (f *Fetcher) loop() {
 		case notification := <-f.notify:
 			// A block was announced, make sure the peer isn't DOSing us
 			count := f.announces[notification.origin] + 1
-			if count > hashLimit {
-				log.Info("Peer exceeded outstanding announces", "peer", notification.origin, "hash-limit", hashLimit)
+			if count > HashLimit {
+				log.Info("Peer exceeded outstanding announces", "peer", notification.origin, "hash-limit", HashLimit)
 				break
 			}
 			// All is well, schedule the announce if block's not yet downloading
--- a/protocol/fetcher/fetcher.go
+++ b/protocol/fetcher/fetcher.go
@@ -310,13 +310,10 @@ func (f *Fetcher) loop() {
 			// Send out all block requests
 			for peer, hashes := range request {
 				if len(hashes) > 0 {
-					list := "["
-					for _, hash := range hashes {
-						list += fmt.Sprintf("%x, ", hash[:4])
-					}
-					list = list[:len(list)-2] + "]"
-
-					log.Debug("fetching", "peer", peer, "hashes", list)
+					// Log a summary rather than the whole list: formatting
+					// every hash costs work proportional to the batch before
+					// the logger decides whether debug output is even on.
+					log.Debug("fetching", "peer", peer, "count", len(hashes), "first-hash", fmt.Sprintf("%x", hashes[0][:4]))
 				}
 				// Create a closure of the fetch and schedule in on a new thread
 				fetcher, hashes := f.fetching[hashes[0]].fetch, hashes
--- a/protocol/handler.go
+++ b/protocol/handler.go
@@ -307,6 +307,11 @@ func (pm *ProtocolManager) handleMsg(p *peer) error {
 			} else if err != nil {
 				return errResp(ErrDecode, "msg %v: %v", msg, err)
 			}
+			// Every requested hash costs a store lookup whether or not the
+			// block exists, so bound the request itself, not only the hits.
+			if len(hashes) >= MaxBlocksRequest {
+				return errResp(ErrMsgTooLarge, "msg %v: more than %d block hashes requested", msg, MaxBlocksRequest)
+			}
 			hashes = append(hashes, hash)
 
 			// Retrieve the requested block, stopping if enough was found
@@ -319,13 +324,7 @@ func (pm *ProtocolManager) handleMsg(p *peer) error {
 		}
 
 		if len(blocks) == 0 && len(hashes) > 0 {
-			list := "["
-			for _, hash := range hashes {
-				list += fmt.Sprintf("%x, ", hash[:4])
-			}
-			list = list[:len(list)-2] + "]"
-
-			log.Debug("no blocks found for requested hashes", "peer-id", p.id, "hashes", list)
+			log.Debug("no blocks found for requested hashes", "peer-id", p.id, "count", len(hashes), "first-hash", fmt.Sprintf("%x", hashes[0][:4]))
 		}
 		return p.SendBlocks(blocks)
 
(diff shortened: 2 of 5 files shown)


edgepillar:
Thanks for this hardening. I reviewed `2d25c88`; the lookup check is placed before the database access, so misses and repeated hashes count toward the 256-lookup bound. Outbound chunking preserves order, and ordinary downloader requests remain within their existing 128-block limit.

I did not find a newly introduced merge-blocking defect in that scoped change. This was a source-level review; I have not independently executed the reported validation.

**1. The deferred fetcher counter issue deserves a separate security follow-up**

The existing announced → fetching transition calls `forgetHash`, decrementing the announcing peer’s counter. Completion or timeout later decrements it again for the fetching entry, without a corresponding increment at the transition.

This permits negative counters and progressively larger accepted announcement sets across repeated lifecycles; the drift is not inherently bounded to a small amount. See the [transition](https://github.com/zenon-network/go-zenon/blob/2d25c886ba31a3a366b424298cdff1b41f7c1519/protocol/fetcher/fetcher.go#L293-L307) and [cleanup accounting](https://github.com/zenon-network/go-zenon/blob/2d25c886ba31a3a366b424298cdff1b41f7c1519/protocol/fetcher/fetcher.go#L468-L487).

The fetcher also constructs the complete debug string using repeated concatenation **before** calling the newly chunked sender. Consequently, chunking does not address that separate CPU/allocation amplification path. See [request logging and dispatch](https://github.com/zenon-network/go-zenon/blob/2d25c886ba31a3a366b424298cdff1b41f7c1519/protocol/fetcher/fetcher.go#L310-L327).

These behaviors predate this PR and the counter issue is explicitly deferred in its description. I am suggesting separate tracking, not treating them as regressions introduced here.

Useful follow-up tests would assert accurate, nonnegative accounting throughout announcement, fetching, completion and timeout, together with a bounded combined retained-state budget. Log formatting should also have an independent bound.

**2. Compatibility and additional coverage**

The documented rejection of older, unsplit, miss-heavy requests exceeding 256 hashes is a deliberate compatibility tradeoff. A mixed-version test would help confirm that behavior.

The existing handler and chunk-order tests are useful. Optional additions would cover malformed RLP/hash encodings, exact-multiple chunk sizes, empty outbound requests, and a write failure on a later chunk.

Overall, this looks like a useful per-request work bound, but it should not be presented as a complete fetcher quota or request-rate defense.

0x3639:
Verified both points against the code.

**1. Counter drift.** Confirmed: the transition into `fetching` releases the announce via `forgetHash` and never re-counts the fetching entry, and the `== 0` delete lets a counter that passes through zero stay negative. A test driving eight hashes through announce, fetch and completion reads -8 afterwards, and after two completed batches of 256 the peer is allowed 264 announces. Fixed in a separate PR as suggested: #95 increments the counter at the transition so announced plus fetching entries equal the counter at every point, floors `forgetHash` at zero, and adds tests over announce, fetching, completion and timeout plus the cross-lifecycle bound.

The fetcher's own "fetching" debug line is bounded in this PR (006d6d2), the same way as the handler's all-miss line: count and first hash prefix instead of the formatted list.

**2. Coverage.** Added in 006d6d2:
- an unsplit request from a peer on an earlier release: answered when `MaxBlockFetch` hits occur within the first 256 hashes, dropped after exactly 256 lookups otherwise (the drop case fails on the unpatched handler);
- malformed payloads (non-list, short hash, long hash, nested list) are rejected at the malformed element with no reply;
- `RequestBlocks` with 0 hashes sends one empty frame as before, and an exact multiple of 256 sends full frames with no trailing empty one; frame counts are now asserted for every size;
- a write failure on a later chunk is returned and no further chunk is attempted, with every write attempt counted so a sender that ignored the error would fail the test.

One pre-existing quirk surfaced while probing malformed input: a payload of an empty list followed by trailing bytes is accepted, because the handler stops at the end of the list and never checks for trailing data. It does not affect the lookup bound and is left as is.

Agreed on framing: this is a per-request work bound, not a fetcher quota or request-rate defense. On compatibility, the drift means an unpatched fetcher can in practice batch more than 256 hashes into one unsplit request after a few completed lifecycles, so "an announcer holds the blocks it announced" is the honest-peer assumption behind the tradeoff, not a guarantee. The counter fix removes the drift on patched nodes.

edgepillar:
Thanks for the follow-up in `006d6d2`. I re-reviewed the changes and the added handler/sender tests.

The fetcher’s debug output now uses a count and first-hash prefix instead of constructing the complete hash list. The added tests meaningfully cover legacy unsplit requests, the 128th hit at the 256-hash boundary, malformed payloads, empty and exact-multiple batches, and stopping after a later chunk’s write failure.

Within this PR’s per-request lookup-bound scope, I consider my logging and coverage concerns addressed in the reviewed source. I found no new blocking issue in the follow-up.

The scope qualifications still matter: the bound limits store lookups, not total payload consumption or request rate, and oversized miss-heavy requests from older, unsplit senders can still cause disconnection. Those are compatibility/resource-policy boundaries, not newly introduced findings.

This was a static review. I did not execute the tests or independently verify successful CI. My accounting/expiry feedback is separate and concerns #95.

openmac0815:

Boundary arithmetic re-checked from the diff: the receiver rejects when appending the 257th hash (`len(hashes) >= MaxBlocksRequest` evaluated before append), so the sender's exact-256 chunks from `RequestBlocks` are accepted — no self-incompatibility. Bounding the request rather than the hits is the right choice since every named hash costs a store lookup. The debug-log summary change also removes per-batch formatting cost proportional to the hash count. Compatible with #95's `HashLimit` export in either merge order.

PR #85, opened 2026-09-05
rpc: bound subscriptions per connection and server-wide

0x3639:
## Summary

The `ledger` subscription API kept every accepted subscription with no ceiling, in two places: the RPC handler's per-connection map and the subscribe server's per-type maps that every matching chain event is fanned out to from the single worker goroutine. The 100-entry install backlog only bounds entries the worker has not drained yet. Closed subscriptions were also only reclaimed when a broadcast happened to visit them, which for address-filtered subscriptions means when a block for that exact address arrives.

This PR adds two bounds, enforced through the existing entry points and released exactly once, and makes reclamation independent of the worker:

- **Per connection (rpc/server):** at most 64 server subscriptions. The slot is reserved in `handleSubscribe` after argument validation and before the notifier is created, and counts calls still in flight, so a batch or concurrent requests cannot exceed it. If the callback returns without creating a subscription the reservation is released right away, so rejected elements of a batch do not count against later elements of the same batch; otherwise `addSubscriptions` converts it into a map entry. Rejections return `ErrTooManySubscriptions` and create no notifier.
- **Server-wide (rpc/api/subscribe):** at most 4096 live subscriptions, counted from acceptance in `subscribe` (under `stopLock`) until `uninstall`, which is idempotent and releases the slot under `subsMu` together with the map entry. Reported as `ErrSubscriptionLimitReached` ahead of the transient backlog check.
- **Reclamation:** each installed subscription gets a watcher goroutine that waits on the RPC subscription's error channel and the connection's closed channel (the same signals `Closed` reports) and uninstalls the entry as soon as either fires. Capacity therefore returns when the client unsubscribes or disconnects, without waiting for the worker or for an event that matches the entry. The subscription maps are guarded by `subsMu`; the worker broadcasts to a snapshot so notification writes never hold the lock. Watchers are bounded by the global limit and exit on `Stop`.

The limits are constants for now; the subscribe server is created by the core without access to the RPC config. Current official clients stay well below them (Syrius opens two subscriptions per connection; the SDK exposes four subscription methods with no automatic address fan-out).

Out of scope, tracked in #96: notification fan-out still runs synchronously on the worker with no slow-consumer policy, and the `Notifier.Notify` doc comment claims a connection close on error that does not happen.

## Tests

- `rpc/server/subscription_test.go` (new): limit reached exactly at 64 and unsubscribe frees a slot; a batch larger than the limit accepts exactly 64; service-side rejections and argument errors consume no budget; a batch of 64 rejections followed by one acceptable request accepts it and uses one slot; a subscription created before the service errors keeps its slot; per-connection budgets are independent and closing a connection releases everything.
- `rpc/api/subscribe/api_internal_test.go`: through `GetSubscribeServer`, `GetSubscribeApi` and an in-process RPC client, exactly 4096 subscriptions are accepted and the next is rejected; an unsubscribed momentum subscription returns capacity; 4096 address-filtered subscriptions whose connections all close return capacity with no matching event; subscribe/unsubscribe churn on one connection totalling 4096 address-filtered subscriptions leaves a fresh client admitted.

All new tests fail on the unpatched code.

## Verification

- `GOWORK=off go vet ./rpc/...`
- `GOWORK=off go test -race -count=3 ./rpc/server/ ./rpc/api/subscribe/`
- `GOWORK=off golangci-lint run ./rpc/...` unchanged from baseline
- `GOWORK=off go test ./...`

### DIFF
--- a/rpc/api/subscribe/api.go
+++ b/rpc/api/subscribe/api.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"errors"
 	"sync"
+	"sync/atomic"
+	"time"
 
 	"github.com/inconshreveable/log15"
 
@@ -18,11 +20,28 @@ import (
 // full; the caller should retry.
 var ErrSubscribeBacklogFull = errors.New("subscribe server backlog is full")
 
+// ErrSubscriptionLimitReached is returned when the server already serves
+// maxSubscriptions live subscriptions across all connections.
+var ErrSubscriptionLimitReached = errors.New("subscribe server subscription limit reached")
+
 const (
 	acChanSize  = 100
 	mChanSize   = 100
 	installSize = 100
+
+	// maxSubscriptions bounds the subscriptions kept installed across all
+	// connections. Every matching chain event is delivered to each of them
+	// from the single worker goroutine, so this also bounds the work one
+	// event can cost. Each connection is separately limited by the RPC
+	// server.
+	maxSubscriptions = 4096
 )
+
+// sweepInterval is how often the worker drops subscriptions whose client has
+// unsubscribed or disconnected. Broadcasts only visit subscriptions their
+// event matches, so without the sweep an address-filtered subscription
+// would stay installed until a block for that address arrives.
+var sweepInterval = 10 * time.Second
 
 var (
 	oneSingleton sync.Mutex
@@ -70,6 +89,13 @@ type Api struct {
 	isStopped bool
 	installCh chan *Subscription // add subscription
 	stopped   chan struct{}
+
+	// live counts subscriptions that hold a slot of maxSubscriptions: those
+	// waiting in installCh plus those installed by the worker. Only subscribe
+	// increments it, serialized by stopLock, and only the worker's uninstall
+	// decrements it, so the check-then-increment in subscribe cannot
+	// overshoot.
+	live atomic.Int64
 }
 type Server struct {
 	*Api
@@ -179,6 +205,8 @@ func (s *Server) work() {
 	defer common.RecoverStack()
 	log.Info("start event loop")
 	defer log.Info("stop event loop")
+	sweep := time.NewTicker(sweepInterval)
+	defer sweep.Stop()
 	for {
 		select {
 		case <-s.stopped:
@@ -187,6 +215,8 @@ func (s *Server) work() {
 			return
 		case sub := <-s.installCh:
 			s.install(sub)
+		case <-sweep.C:
+			s.sweep()
 		case momentums := <-s.mCh:
 			s.broadcastMomentums(momentums)
 		case blocks := <-s.acCh:
@@ -205,8 +235,32 @@ func (s *Server) install(subscription *Subscription) {
 	s.subscriptions[subscription.options.subscriptionType][subscription.rpc.ID] = subscription
 }
 func (s *Server) uninstall(subscription *Subscription) {
+	installed := s.subscriptions[subscription.options.subscriptionType]
+	if _, ok := installed[subscription.rpc.ID]; !ok {
+		return
+	}
 	s.log.Info("uninstall", "id", subscription.rpc.ID)
-	delete(s.subscriptions[subscription.options.subscriptionType], subscription.rpc.ID)
+	delete(installed, subscription.rpc.ID)
+	// Released exactly once, together with the map entry the slot was held for.
+	s.live.Add(-1)
+}
+
+// sweep uninstalls every subscription whose client has unsubscribed or
+// disconnected, regardless of whether an event for it has arrived.
+func (s *Server) sweep() {
+	startTime := common.Clock.Now()
+	stats := &BroadcastStats{}
+	for _, installed := range s.subscriptions {
+		for _, subscription := range installed {
+			if subscription.Closed() {
+				stats.NumUninstalls += 1
+				s.uninstall(subscription)
+			}
+		}
+	}
+	if stats.NumUninstalls > 0 {
+		s.log.Info("finish sweeping subscriptions", "elapsed", common.Clock.Now().Sub(startTime), "stats", stats)
+	}
 }
 func (s *Server) broadcast(subscription *Subscription, data interface{}, stats *BroadcastStats) {
 	if subscription.Closed() {
@@ -300,10 +354,17 @@ func (s *Api) subscribe(ctx context.Context, options *subscriptionOptions) (*rpc
 	// client never learns and cannot unsubscribe. stopLock serializes
 	// producers and the worker only drains, so once there is room the send
 	// below cannot block.
+	// The slot is claimed here, before the entry is handed to the worker, so
+	// the count covers queued and installed subscriptions alike; the worker
+	// returns it when it uninstalls the entry.
+	if s.live.Load() >= maxSubscriptions {
+		return nil, ErrSubscriptionLimitReached
+	}
 	if len(s.installCh) == cap(s.installCh) {
 		return nil, ErrSubscribeBacklogFull
 	}
 	subscription := NewSubscription(notifier, options)
+	s.live.Add(1)
 	s.installCh <- subscription
 	return subscription.rpc, nil
 }
--- a/rpc/api/subscribe/api.go
+++ b/rpc/api/subscribe/api.go
@@ -5,7 +5,6 @@ import (
 	"errors"
 	"sync"
 	"sync/atomic"
-	"time"
 
 	"github.com/inconshreveable/log15"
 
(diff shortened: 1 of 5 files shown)


edgepillar:
Thanks for bounding subscription retention. I reviewed `e7de926`; accounting for pending callbacks and serializing admission appear to support the stated 64-per-connection and 4096-global bounds. I did not find a demonstrated limit-bypass race or negative-counter path.

This is a source-level review; I have not run a runtime reproducer or independently reproduced the reported tests.

**1. Failed callbacks retain reservations until the whole batch finishes**

Could reservations for callbacks that return without creating a subscription be released before processing the next batch element?

Currently, `addSubscriptions` releases them only after the entire batch. A batch with 64 service-side rejections followed by an otherwise valid subscription request therefore rejects the final request with `ErrTooManySubscriptions`, despite having created no subscriptions. See [batch ordering](https://github.com/zenon-network/go-zenon/blob/e7de926c6f55764c6ad69a9be3757ea104f8aa60/rpc/server/handler.go#L119-L132) and [reservation conversion](https://github.com/zenon-network/go-zenon/blob/e7de926c6f55764c6ad69a9be3757ea104f8aa60/rpc/server/handler.go#L200-L228).

This is a minor correctness edge, not a demonstrated security exploit. For the ledger service, a practical instance would require a transient rejection condition to clear during the batch. A fake service can test the accounting deterministically.

The correction should preserve reservations for genuinely pending callbacks and created subscriptions.

**2. Closed subscriptions can continue consuming the global admission budget**

Unsubscribe immediately frees the connection’s slot, but the global slot remains occupied until the worker removes the retained entry. For an address without matching events, one connection can accumulate closed subscriptions against the global limit between sweeps while remaining below its per-connection limit. See [unsubscribe](https://github.com/zenon-network/go-zenon/blob/e7de926c6f55764c6ad69a9be3757ea104f8aa60/rpc/server/handler.go#L432-L443) and [global admission](https://github.com/zenon-network/go-zenon/blob/e7de926c6f55764c6ad69a9be3757ea104f8aa60/rpc/api/subscribe/api.go#L357-L368).

A controlled-tick test could exercise subscribe/unsubscribe churn, then check whether a fresh client is rejected solely because closed entries remain retained. I have not measured production throughput or demonstrated a live attack.

The previous implementation retained these entries too, so this is useful memory hardening. The question is whether delayed reclamation causing admission starvation is an accepted operational tradeoff.

Any earlier capacity release should remain coupled to removal or retirement of the retained entry. Merely decrementing the counter while leaving entries installed would undermine the memory bound.

**3. The sweep interval does not bound capacity-return latency**

Cleanup shares the worker with synchronous notification delivery. Slow writes or accumulated fanout can therefore delay both sweeps and installation; a 10-second ticker does not guarantee reclamation within 10 seconds. Individual write deadlines do not bound the entire fanout or time waiting for the encoder lock. See the [worker](https://github.com/zenon-network/go-zenon/blob/e7de926c6f55764c6ad69a9be3757ea104f8aa60/rpc/api/subscribe/api.go#L203-L224) and [write path](https://github.com/zenon-network/go-zenon/blob/e7de926c6f55764c6ad69a9be3757ea104f8aa60/rpc/server/json.go#L223-L232).

There is also an existing lifecycle mismatch: `Notifier.Notify` documents connection closure on error, but the inspected send path returns the write error and the subscribe wrapper logs it without closing the codec. Reclamation can still depend on subsequent read-side closure or explicit unsubscribe. See [notification error handling](https://github.com/zenon-network/go-zenon/blob/e7de926c6f55764c6ad69a9be3757ea104f8aa60/rpc/api/subscribe/subscription.go#L70-L93).

These are pre-existing delivery/lifecycle limitations, suitable for explicit follow-up rather than automatically blocking this PR.

Controlled blocking/erroring writers, barrier-paused concurrent callbacks, and unsubscribe/sweep/broadcast interleavings would provide deterministic regression coverage. An overall deadline on backlog-retry test helpers would also prevent a stalled worker from hanging until the suite timeout.

Overall, the bounds are valuable; I would tighten the mixed-batch behavior and explicitly define the reclamation and slow-consumer guarantees.

0x3639:
Addressed in e01da4f.

**1. Mixed batches.** Confirmed. `handleSubscribe` now checks the notifier as soon as the callback returns and releases the reservation immediately when no subscription was created; only notifiers holding a subscription are collected by `addSubscriptions`. Reservations for calls still in flight and for created subscriptions are untouched. Test: `TestRejectedSubscriptionInBatchFreesSlotForLaterElement` sends 64 service rejections followed by one acceptable request in one batch, asserts the last one succeeds and that exactly one slot is in use afterwards. It failed on e7de926 with `ErrTooManySubscriptions` on the last element.

**2. Closed entries holding the global budget.** Not accepted as a tradeoff. One connection can cycle through 4096 subscribe/unsubscribe pairs well inside the sweep interval, so the global cap was effectively holdable by a single client. Reclamation no longer goes through the worker: each installed entry gets a watcher goroutine that waits on the RPC subscription's error channel and the connection's closed channel, the same two signals `Closed` reports, and uninstalls the entry the moment either fires. The counter is still decremented only inside `uninstall`, under `subsMu` together with the map delete, so release stays coupled to removal and cannot happen twice. The maps are now guarded by `subsMu`, and the worker broadcasts to a snapshot so notification writes never hold the lock. Watchers are bounded by the global limit, take their own copies of the notifier and RPC subscription so they never touch `Subscription.notifier` (still worker-owned), and exit on `Stop`. The 10 s sweep is removed since the watcher subsumes it. Test: `TestUnsubscribeChurnDoesNotStarveGlobalCapacity` does the churn on one connection with address-filtered subscriptions and no matching event, then asserts a fresh client is admitted. It failed on e7de926 after the 10 s wait.

**3. Fan-out latency and the `Notify` doc mismatch.** Agreed these are pre-existing and out of scope here. Opened #96 with the details, the suggested slow-consumer policy options, and the deterministic test coverage you listed (controllable writers, interleavings, an overall deadline on the backlog-retry helpers). With the watcher change, a stalled worker now delays installs and delivery but no longer delays the return of global slots.

Verified with `go test -race -count=3` on both packages, lint unchanged from baseline, and the full suite. PR description updated to match.

edgepillar:
Thanks for the follow-up in `e01da4f`. The immediate release of reservations for callbacks that create no subscription addresses the mixed-batch concern.

This remains a source-only review; I have not executed a reproducer or independently verified CI.

**P1 — Capture the watcher’s signals before launching its goroutine**

`install` launches `go s.watch(subscription)`, but the notifier copy is taken inside that new goroutine. Meanwhile, a matching broadcast can call `Subscription.Closed()`, which clears `subscription.notifier` after unsubscribe or disconnect. The copy therefore still races with worker-side cleanup. [Launch and read](https://github.com/zenon-network/go-zenon/blob/e01da4f902694ce6d1e1172374206ac60aa4d518/rpc/api/subscribe/api.go#L230-L253), [mutation](https://github.com/zenon-network/go-zenon/blob/e01da4f902694ce6d1e1172374206ac60aa4d518/rpc/api/subscribe/subscription.go#L80-L93).

If cleanup wins before the copy, the watcher evaluates `notifier.Closed()` with a nil receiver. That method dereferences the notifier, and the watcher has no panic recovery. An already-ready subscription-error or stopped channel does not prevent evaluation of the other `select` operands. [Dereference](https://github.com/zenon-network/go-zenon/blob/e01da4f902694ce6d1e1172374206ac60aa4d518/rpc/server/subscription.go#L151-L155).

Please capture the RPC error and connection-closed channels synchronously in `install`, before launching the goroutine, and pass those immutable signals to the watcher. A nil check alone would not remove the unsynchronized access.

A barrier-controlled regression should delay watcher execution, unsubscribe or disconnect, complete a matching broadcast that clears the worker-owned field, then release the watcher. Check no panic, exactly-once removal/quota release, and termination on `Stop`.

Separately, the churn test does not establish worker-independent reclamation: it permits a ten-second wait, during which the previous ten-second sweep could also reclaim capacity. Parking the worker in a controlled notification write while checking removal of another installed subscription would distinguish the two implementations.

The P1 is a source-supported, timing-dependent process-panic path—not an independently demonstrated remote exploit—and is separate from #96.

openmac0815:

Counter pairing verified: `live` is incremented only in `subscribe` (serialized by stopLock) and decremented exactly once in `uninstall` under `subsMu` together with the map entry — no overshoot, no double release. The watcher-goroutine reclamation makes capacity return independent of the broadcast worker and of whether a matching event ever arrives, which fixes the address-filtered-subscription leak. Snapshot-broadcast keeps slow-client writes out of `subsMu`. The per-connection reservation-in-`handleSubscribe` before notifier creation correctly closes the batch/concurrent overshoot window.

PR #86, opened 2026-09-05
p2p/discover: admit inbound bonds before starting them

0x3639:
## Summary

An unsolicited discovery ping made the handler start a bonding goroutine unconditionally, and `bond` registered the identity in the bonding map before `pingpong` waited for one of the 16 bonding slots. The slot limit therefore only bounded active ping/pong exchanges: goroutines and map entries waiting for a slot grew with the arrival rate, and since a remote that never answers holds a slot for the 500ms response timeout, a modest sustained rate of valid pings from fresh identities kept that backlog growing for as long as the traffic lasted.

Changes:

- **Admission before spawn** (`p2p/discover/udp.go`): the ping handler takes a slot of a new inbound bonding budget (`maxInboundBonds`, 64, covering waiting and active inbound bonds) with a non-blocking send before starting the goroutine; the goroutine returns it on exit. When the budget is exhausted the ping is still answered with a pong, only the follow-up bond from our side is skipped and logged at debug level; the remote can ping again later. Outbound bonding during lookups is unchanged.
- **Close-aware slot wait** (`p2p/discover/table.go`): `pingpong` stops waiting for a bonding slot when the table is closing and fails the process with `errClosed`, so queued processes exit at shutdown instead of parking.
- **Helper goroutine removed**: it raced on its `ok` flag with the deferred slot release (`go test -race` reports it on the unpatched code when `Close` runs during a bond) and could consume a slot from the pool for the duration of a bond. The slot is now returned unconditionally; the pool holds exactly `cap(bondslots)` tokens and the process holds one, so the send cannot block.

## Tests

`p2p/discover/udp_test.go` (new; the package had no tests). A datagram fake records writes by destination while packets are signed with `encodePacket` and delivered through `udp.handlePacket`, so the real decoder and handler run.

- 114 distinct identities that never answer: every ping gets a pong, at most 64 bonding processes are registered, goroutine growth stays within the budget.
- 192 pings from one identity: a single bonding process, bounded goroutines.
- After the admitted bonds time out, a later sender is bonded again.
- Closing the table with the budget full lets every transport goroutine exit.

On the unpatched code the first two tests observe 114 bonding processes and 192 new goroutines, and the close test times out with goroutines still parked.

## Verification

- `GOWORK=off go vet ./p2p/discover/`
- `GOWORK=off go test -race -count=3 ./p2p/discover/`
- `GOWORK=off go test ./...`

### DIFF
--- a/p2p/discover/table.go
+++ b/p2p/discover/table.go
@@ -44,6 +44,9 @@ const (
 	nBuckets   = hashBits + 1 // Number of buckets
 
 	maxBondingPingPongs = 16
+	// maxInboundBonds bounds the bonding processes started for unsolicited
+	// pings, whether they are waiting for a bonding slot or holding one.
+	maxInboundBonds     = 64
 	maxFindnodeFailures = 5
 )
 
@@ -58,6 +61,10 @@ type Table struct {
 	bondmu    sync.Mutex
 	bonding   map[NodeID]*bondproc
 	bondslots chan struct{} // limits total number of active bonding processes
+	// inboundBonds is a counting semaphore for bonds started by unsolicited
+	// pings; a slot is taken before the goroutine is started and released
+	// when it exits, so work is admitted before it is queued anywhere.
+	inboundBonds chan struct{}
 
 	nodeAddedHook func(*Node) // for testing
 
@@ -105,6 +112,8 @@ func newTable(t transport, ourID NodeID, ourAddr *net.UDPAddr, nodeDBPath string
 		closing:   make(chan struct{}),
 		bonding:   make(map[NodeID]*bondproc),
 		bondslots: make(chan struct{}, maxBondingPingPongs),
+
+		inboundBonds: make(chan struct{}, maxInboundBonds),
 	}
 	for i := 0; i < cap(tab.bondslots); i++ {
 		tab.bondslots <- struct{}{}
@@ -433,22 +442,19 @@ func (tab *Table) bond(pinged bool, id NodeID, addr *net.UDPAddr, tcpPort uint16
 }
 
 func (tab *Table) pingpong(w *bondproc, pinged bool, id NodeID, addr *net.UDPAddr, tcpPort uint16) {
-	// Request a bonding slot to limit network usage
-	<-tab.bondslots
-	ok := true
-	go func() {
-		select {
-		case <-w.done:
-		case <-tab.bondslots:
-		case <-tab.closing:
-			ok = false
-		}
-	}()
-	defer func() {
-		if ok {
-			tab.bondslots <- struct{}{}
-		}
-	}()
+	// Request a bonding slot to limit network usage. Waiting also ends at
+	// shutdown so queued processes exit at once instead of each taking a
+	// slot and failing through the closed transport in turn.
+	select {
+	case <-tab.bondslots:
+	case <-tab.closing:
+		w.err = errClosed
+		close(w.done)
+		return
+	}
+	// The slot is always returned: there are exactly cap(bondslots) tokens
+	// and this process holds one, so the send cannot block.
+	defer func() { tab.bondslots <- struct{}{} }()
 
 	// Ping the remote side and wait for a pong
 	if w.err = tab.ping(id, addr); w.err != nil {
--- a/p2p/discover/table.go
+++ b/p2p/discover/table.go
@@ -61,10 +61,13 @@ type Table struct {
 	bondmu    sync.Mutex
 	bonding   map[NodeID]*bondproc
 	bondslots chan struct{} // limits total number of active bonding processes
-	// inboundBonds is a counting semaphore for bonds started by unsolicited
-	// pings; a slot is taken before the goroutine is started and released
-	// when it exits, so work is admitted before it is queued anywhere.
-	inboundBonds chan struct{}
+	// inbound holds the identities with a bond started by an unsolicited
+	// ping, whether waiting for a bonding slot or holding one. An identity
+	// is admitted before its goroutine starts and removed when it exits, so
+	// work is bounded before it is queued anywhere, and an identity that is
+	// already in flight is not admitted again: its pings coalesce onto the
+	// existing process instead of each taking a permit.
+	inbound map[NodeID]struct{}
 
 	nodeAddedHook func(*Node) // for testing
 
@@ -112,8 +115,7 @@ func newTable(t transport, ourID NodeID, ourAddr *net.UDPAddr, nodeDBPath string
 		closing:   make(chan struct{}),
 		bonding:   make(map[NodeID]*bondproc),
 		bondslots: make(chan struct{}, maxBondingPingPongs),
-
-		inboundBonds: make(chan struct{}, maxInboundBonds),
+		inbound:   make(map[NodeID]struct{}),
 	}
 	for i := 0; i < cap(tab.bondslots); i++ {
 		tab.bondslots <- struct{}{}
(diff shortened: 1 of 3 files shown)


edgepillar:
Thanks for moving admission ahead of goroutine creation. I reviewed `6454818`; the pre-spawn check and removal of the extra semaphore consumer look sound. The revised code bounds inbound workers, including duplicate-identity waiters, and returns each acquired exchange slot exactly once.

This is a source-level review; I have not independently executed the reported race tests or full suite.

**1. Failed bonds can leave metadata that expiration does not reclaim**

`Table.ping` writes `:discover:lastping` before receiving a pong. If the exchange fails, `pingpong` returns before creating the corresponding `:discover` node record. However, `expireNodes` only processes node records, so metadata-only identities are skipped even after their timestamps become old. See [the write-before-response path](https://github.com/zenon-network/go-zenon/blob/6454818a1b96b8068f8e5a0c308624c09df242ac/p2p/discover/table.go#L495-L505) and [expiration filtering](https://github.com/zenon-network/go-zenon/blob/6454818a1b96b8068f8e5a0c308624c09df242ac/p2p/discover/database.go#L249-L284).

Sustained fresh, unanswered identities can therefore continue increasing database storage despite the concurrency bound.

This behavior already exists in the base branch; it is not introduced by this PR. Could we track expiration of metadata-only identities, or another bounded retention policy that preserves legitimate known-peer state?

A focused regression could create stale `lastping` records without node records, invoke expiration directly, and verify cleanup alongside a fresh bonded-node control.

**2. Repeated pings from one identity can occupy all inbound permits**

Admission occurs before the same-ID check in `bond`. One active exchange and its duplicate waiters can therefore occupy all 64 permits, causing other inbound identities to lose their follow-up bond attempts even when additional exchange slots are available. See [admission](https://github.com/zenon-network/go-zenon/blob/6454818a1b96b8068f8e5a0c308624c09df242ac/p2p/discover/udp.go#L538-L554) and [same-ID waiting](https://github.com/zenon-network/go-zenon/blob/6454818a1b96b8068f8e5a0c308624c09df242ac/p2p/discover/table.go#L406-L411).

The total worker count remains bounded and the permits are released when the exchange finishes. This is an admission-fairness limitation, not an unbounded leak or a claim that outbound discovery is categorically blocked.

Would it make sense to coalesce redundant same-ID inbound work before consuming global capacity, using synchronized admission state?

**3. Strengthen deterministic concurrency coverage**

The fake transport is useful, but fixed sleeps combined with process-wide `runtime.NumGoroutine()` measurements make some observations dependent on scheduling and timeout progress.

Barrier-controlled tests could directly verify:

- The exact admission boundary and rejection beyond it.
- A different identity making progress while repeated same-ID pings arrive.
- Return of all exchange slots across repeated success and failure.
- Cancellation of a worker known to be waiting for a slot.
- Mixed inbound/outbound progress under saturation.

`Table.Close` also does not itself join every worker; tests should distinguish eventual completion after closure from synchronous shutdown.

These observations do not undermine the scoped backlog fix. They identify remaining retention and fairness limits. Packet-processing CPU and pong traffic still occur before admission, and this change covers legacy discovery rather than the separate libp2p discovery path.

0x3639:
Verified all three points against `6454818` and addressed them as follows.

**1. Metadata-only entries never expire.** Confirmed pre-existing: `expireNodes` only considered node-record keys, so a failed exchange left its `lastping` key permanently. Since it is base-branch behaviour and a different mechanism, it is fixed separately in #97: expiration now considers an identity whenever any discovery field is stored for it, under the unchanged last-pong rule. Nothing reads the ping time back, so no known-peer state is lost. The regression test there matches the shape you described: a stale ping-only identity alongside a fresh bonded control and a stale bonded control.

**2. One identity can hold every inbound permit.** Confirmed, fixed in `0a560f3`. Admission is now a set of in-flight inbound identities guarded by the bonding mutex; an identity already in flight is not admitted again. Dropping the duplicate is lossless, since the waiter it would have created only re-does the bucket update the in-flight process performs anyway. Permits are therefore consumed per distinct identity. `TestRepeatedIdentityDoesNotStarveOthers` sends 192 pings from one identity, checks that one permit is held, and shows a different identity admitted and bonded while the first is still in flight. On the previous admission code it fails with 64 permits held.

**3. Deterministic coverage.** Also in `0a560f3`. The fake socket can now hold outbound pings at the write until released, which parks every admitted bond at a known point with its permit and slot taken; senders can also answer the ping addressed to them. The fixed sleeps and `runtime.NumGoroutine()` checks are gone. Coverage now includes:

- exact admission boundary, checked synchronously after each ping, and rejection beyond it (`TestInboundBondingIsBounded`)
- a different identity progressing during repeated same-id pings (above)
- return of every permit and slot across mixed success and timeout, with the bonded set checked afterwards (`TestInboundBondingBudgetIsReturned`)
- an outbound bond completing with the inbound budget and every slot saturated, without taking an inbound permit (`TestOutboundBondProgressesUnderInboundSaturation`)
- workers exiting after close, with the test comment stating that `Close` does not itself join them (`TestQueuedBondsExitOnClose`)

The one timing dependency left is the 500 ms response timeout on the failure paths, which is inherent to what those tests observe; the assertions on it are eventual with generous deadlines, not fixed sleeps.

Verified with `go vet`, `go test -race -count=3 ./p2p/discover/`, and the full suite.

edgepillar:
Thanks for addressing these points in `0a560f3` and separating metadata retention into #97. This remains a source-only review; I have not independently executed the reported tests.

The identity-set admission addresses the fairness concern: duplicate identities cannot consume additional inbound permits, and release remains deferred until the worker exits. I found no new production defect in that correction.

**P3 — Successful exchanges still depend on timeout scheduling**

The fake socket holds pings inside `WriteToUDP`, but `udp.ping` registers its pending response before sending. The pending loop starts the 500 ms deadline when accepting that request. Holding the write therefore does not hold the deadline: an intended-success exchange can expire before release, making the subsequent pong unsolicited. A longer assertion deadline cannot recover the expired request. [Write gate](https://github.com/zenon-network/go-zenon/blob/0a560f3fd4cde5f2ad6dab35230145b40c59c1d0/p2p/discover/udp_test.go#L42-L49), [registration](https://github.com/zenon-network/go-zenon/blob/0a560f3fd4cde5f2ad6dab35230145b40c59c1d0/p2p/discover/udp.go#L264-L270), [deadline](https://github.com/zenon-network/go-zenon/blob/0a560f3fd4cde5f2ad6dab35230145b40c59c1d0/p2p/discover/udp.go#L364-L367).

The mixed-budget test also answers successful senders serially by index, although slot acquisition order is unspecified. Waiting for one queued sender can let another active exchange expire. [Answer loop](https://github.com/zenon-network/go-zenon/blob/0a560f3fd4cde5f2ad6dab35230145b40c59c1d0/p2p/discover/udp_test.go#L297-L300).

Could success cases use responders driven by actual outbound-request events, together with controlled timeout progression? Dedicated failure cases can retain real timeout coverage.

One smaller coverage improvement: confirm the outbound identity has entered `bonding` before releasing saturation. Otherwise its goroutine may start only after release, weakening the intended assertion. [Outbound test](https://github.com/zenon-network/go-zenon/blob/0a560f3fd4cde5f2ad6dab35230145b40c59c1d0/p2p/discover/udp_test.go#L346-L356).

These are test-reliability and coverage points, not objections to the production coalescing fix.

openmac0815:

Admit-before-spawn with the 64-permit inbound budget closes the real gap: previously the 16 bonding slots only bounded active ping/pongs while goroutines and map entries grew with arrival rate. Verified: the pong is still sent when the budget is exhausted (no protocol regression for the remote), per-identity coalescing prevents one remote from burning multiple permits, `releaseInbound` runs via defer on goroutine exit (no leak path), and the close-aware slot wait ends queued processes at shutdown instead of failing them through a closed transport one by one. The slot-return send can't block (token-count invariant holds).

PR #87, opened 2026-09-05
node: honor the RPC enable flags at startup

0x3639:
## Summary

`RPCConfig` carries `EnableHTTP` and `EnableWS` next to the host and port of each protocol, and the CLI maps `--http` and `--ws` onto those booleans, but `startRPC` only ever looked at the host strings. An operator who set an enable flag to `false` while leaving the host configured (the shipped defaults are `0.0.0.0` for both) still got that protocol's handler installed and its socket bound. The only way to actually turn a protocol off was the undocumented one of emptying its host.

Changes (`node/rpc.go`):

- A protocol is configured only when its flag is set **and** its host is non-empty. Otherwise no listen address is set, no handler is installed and no listener is started, so `listenAddr` reports nothing for it and the port stays free.
- When a flag is `false` but a host is configured, one info line says the protocol was disabled by configuration, so operators who relied on the old behavior see why it no longer listens.
- `wsServerForPort` takes the effective HTTP state instead of re-deriving it from the host string: WebSocket shares the HTTP server only when HTTP is actually enabled on the same port. With HTTP disabled it previously reached the dedicated server only because the HTTP server's port was still zero.

Config files that omit the flags keep the `true` defaults, and the CLI only writes a flag when `--http` / `--ws` is passed, so existing deployments are unaffected unless they explicitly set a flag to `false`, which is the intended change.

## Tests

`node/rpc_test.go` (new): a node is assembled with only the state `startRPC` needs and started on loopback with ephemeral ports.

- All four combinations of the two flags with both hosts configured: an enabled protocol has a bound socket and answers a JSON-RPC call or a WebSocket handshake; a disabled one has no listener and no endpoint, and the HTTP listener does not upgrade WebSocket connections when only HTTP is enabled.
- The empty-host behavior is unchanged.
- A disabled protocol leaves its configured port free.

Three of the four combinations and the port test fail on the unpatched code.

## Verification

- `GOWORK=off go vet ./node/`
- `GOWORK=off go test -race -count=3 ./node/`
- `GOWORK=off go test ./...`

### DIFF
--- a/node/rpc.go
+++ b/node/rpc.go
@@ -4,8 +4,19 @@ package node
 // startup. It's not meant to be called at any time afterwards as it makes certain
 // assumptions about the state of the node.
 func (node *Node) startRPC() error {
+	// A protocol runs only when its enable flag is set and a host is
+	// configured: the flag is the policy, the host says where to listen.
+	httpEnabled := node.config.RPC.EnableHTTP && node.config.RPC.HTTPHost != ""
+	wsEnabled := node.config.RPC.EnableWS && node.config.RPC.WSHost != ""
+	if !node.config.RPC.EnableHTTP && node.config.RPC.HTTPHost != "" {
+		log.Info("HTTP-RPC server disabled by configuration", "ignored-host", node.config.RPC.HTTPHost)
+	}
+	if !node.config.RPC.EnableWS && node.config.RPC.WSHost != "" {
+		log.Info("WS-RPC server disabled by configuration", "ignored-host", node.config.RPC.WSHost)
+	}
+
 	// Configure HTTP.
-	if node.config.RPC.HTTPHost != "" {
+	if httpEnabled {
 		config := httpConfig{
 			CorsAllowedOrigins: node.config.RPC.HTTPCors,
 			Vhosts:             node.config.RPC.HTTPVirtualHosts,
@@ -21,8 +32,8 @@ func (node *Node) startRPC() error {
 	}
 
 	// Configure WebSocket.
-	if node.config.RPC.WSHost != "" {
-		server := node.wsServerForPort(node.config.RPC.WSPort)
+	if wsEnabled {
+		server := node.wsServerForPort(httpEnabled, node.config.RPC.WSPort)
 		config := wsConfig{
 			Modules: node.config.RPC.Endpoints,
 			Origins: node.config.RPC.WSOrigins,
@@ -42,8 +53,11 @@ func (node *Node) startRPC() error {
 	return node.ws.start()
 }
 
-func (node *Node) wsServerForPort(port int) *httpServer {
-	if node.config.RPC.HTTPHost == "" || node.http.port == port {
+// wsServerForPort returns the server WebSocket should be enabled on: the HTTP
+// server when it is enabled and listens on the same port, otherwise the
+// dedicated WebSocket server.
+func (node *Node) wsServerForPort(httpEnabled bool, port int) *httpServer {
+	if httpEnabled && node.http.port == port {
 		return node.http
 	}
 	return node.ws
--- a/node/rpc_test.go
+++ b/node/rpc_test.go
@@ -1,0 +1,167 @@
+package node
+
+import (
+	"net"
+	"net/http"
+	"strconv"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/gorilla/websocket"
+
+	rpc "github.com/zenon-network/go-zenon/rpc/server"
+)
+
+// probeService is the only API registered by these tests; it exists so a
+// successful JSON-RPC round trip proves the handler is live.
+type probeService struct{}
+
+func (probeService) Ping() string { return "pong" }
+
+// newRPCTestNode builds the node state startRPC works on, without the
+// wallet, chain or p2p server.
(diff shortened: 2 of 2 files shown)


edgepillar:
Reviewed at `4e748b1`. This is a source-based review; I have not independently executed the reported tests.

The effective `EnableHTTP && HTTPHost != ""` / `EnableWS && WSHost != ""` checks look correct. Configuration starts from the enabled defaults, and CLI overrides apply only when explicitly set, so omitted flags retain the intended behavior.

I suggest strengthening the regression coverage and tracking two adjacent, pre-existing defects separately:

1. **Same-port sharing can override the intended bind interface.** When HTTP and WebSocket use the same port, `wsServerForPort` selects the HTTP server without comparing hosts. WebSocket setup then replaces that server's listen address. Consequently, loopback HTTP plus wildcard WebSocket on the same port also exposes HTTP on the wildcard interface. This predates this PR; honoring the enable flags does not introduce it. Please define a policy for conflicting hosts—compatible sharing or an explicit configuration error—rather than silently widening the bind. [Server selection](https://github.com/zenon-network/go-zenon/blob/4e748b1f44348fd67a6dbcae673571c334da99d7/node/rpc.go#L32-L63).

2. **WebSocket shutdown selects the wrong handler.** In `httpServer.doStop`, both `httpHandler` and `wsHandler` are loaded from `h.httpHandler`. The actual WebSocket RPC server therefore misses its `Stop()` call, and established upgraded connections can survive listener shutdown while the process remains alive. This is also pre-existing and outside the patch. A focused follow-up should retain a live WebSocket connection across `stopRPC` and verify closure for both WS-only and shared listeners. [Handler selection](https://github.com/zenon-network/go-zenon/blob/4e748b1f44348fd67a6dbcae673571c334da99d7/node/rpcstack.go#L236-L258).

For this PR's tests:

- Decode the HTTP response and assert the expected ID, `"pong"` result and absence of an RPC error; HTTP 200 alone is insufficient.
- In the WS-only case, also verify that ordinary HTTP RPC is unavailable on the WebSocket listener.
- Cover separate ports, mixed empty/nonempty hosts, and omitted versus explicitly false configuration flags. Both-enabled tests currently use equal ephemeral ports, exercising sharing.
- The disabled-port test can keep its initial listener open while starting disabled RPC, avoiding the release/rebind race.

I did not find a newly introduced security defect in the flag correction itself. The two existing defects above should be distinguished from this PR's test-strengthening suggestions.

0x3639:
Pushed 2db04a0 with the test changes, and split the two pre-existing defects out:

- **Shutdown handler selection**: #98. One-line fix plus a regression test that holds a live WebSocket connection across `stop` for both a WS-only server and a shared listener. Both subtests time out before the fix.
- **Same-port host override**: #99. Filed as an issue with the suggested policy (configuration error when the ports match but the hosts differ). Left out of this PR since it changes accepted configurations.

On the test items:

- Every served protocol is now checked with a decoded JSON-RPC round trip: request id, `"pong"` result, no error field. WebSocket sends the request over the upgraded connection rather than only completing the handshake.
- WS-only asserts plain HTTP JSON-RPC gets a 404 on the WS listener; HTTP-only asserts the HTTP listener refuses the upgrade.
- Added a separate-ports case so the dedicated WS server with HTTP also enabled is exercised. The both-enabled case in the flag table keeps equal ports and now says so, since the shared path is worth pinning too.
- Empty-host is a table covering both-empty and each side empty on its own.
- The disabled-port test holds its probe listener for the whole of `startRPC` instead of releasing and rebinding.

On omitted versus explicitly false flags: at the node layer `RPCConfig` holds plain bools, so the two are the same value by the time `startRPC` runs. The distinction lives in `app/config.go`, where JSON unmarshals over `DefaultNodeConfig` and CLI flags apply only under `ctx.IsSet`, which is the behaviour you confirmed in your opening paragraph. A test for that belongs with the app config loader rather than here; happy to add one there in a separate change if you think it is worth pinning.

edgepillar:
Thanks for the follow-up in `2db04a0` and for separating the adjacent defects into #98 and #99.

The revised tests substantially address my coverage suggestions: decoded HTTP/WebSocket responses, rejection of the disabled protocol, separate-port and mixed-empty-host cases, and keeping the disabled-port probe bound during startup. [Updated tests](https://github.com/zenon-network/go-zenon/blob/2db04a0300936b5e4584ec3d5f10e7769d74aadb/node/rpc_test.go#L184-L305).

I agree that omitted versus explicitly false flags should be tested at the application configuration-loading layer; the node layer’s plain booleans cannot distinguish them.

I found no new production issue in the follow-up. The conflicting-host behavior is now tracked, not resolved, and should remain separate from this PR’s enable-flag correction.

This was a source-only review; I did not execute the tests or independently verify CI.

openmac0815:

Confirmed the bug: `startRPC` only looked at host strings, so `EnableHTTP: false` / `EnableWS: false` with a configured host (the shipped `0.0.0.0` defaults) still bound the socket. The fix gates on flag AND host, logs an info line when a configured host is ignored (good operator UX for the behavior change), and `wsServerForPort` now takes the effective HTTP state so WS shares the HTTP server only when HTTP is actually enabled on that port. Complementary to #88 (either order).

PR #88, opened 2026-09-05
node: serve RPC to local clients only by default

0x3639:
## Summary

The shipped defaults bound both JSON-RPC servers to `0.0.0.0`, granted cross-origin access to every browser origin (`HTTPCors` and `WSOrigins` `"*"`), and registered every public namespace, so a node started without a config file served the whole public RPC surface to any host with a network path and to any website the operator's browser visited. The same defaults also rejected a plain `http://localhost:35997` request with 403, because the empty virtual-host list only admits IP-literal `Host` headers.

Changes:

- **Local-only defaults** (`node/defaults.go`): both servers bind to `127.0.0.1`, the virtual-host list admits `localhost` (IP literals were always admitted), and no CORS or WebSocket origin list is set. With an empty list the CORS middleware is not installed, so no cross-origin grant is issued, and the WebSocket validator falls back to admitting localhost origins and non-browser clients, which keeps wallets and local tools working unchanged. HTTP and WebSocket stay enabled; ports are unchanged.
- **Startup warning** (`node/rpc_exposure.go`, called from `Node.Start` after `startRPC`): one warning per live listener bound to a non-loopback address and per wildcard origin list installed on it. It is derived from the servers' actual state (listener address and installed handlers), not from the configuration, so it reports exactly what is bound regardless of how the enable flags are interpreted. The defaults are silent. `startRPC` itself is untouched; this merges cleanly with #87 in either order.
- **README**: a paragraph on the defaults and how to expose an endpoint.

## Compatibility

- `config.json` files that set the hosts and origins keep working unchanged. The devnet role configs and `devnet-keygen` set them explicitly.
- A node that relied on the bare defaults for remote access must now set `RPC.HTTPHost` / `RPC.WSHost` (or `--http-addr` / `--ws-addr`).
- A `config.json` that sets a host but omits `HTTPCors` / `WSOrigins` no longer inherits the wildcard for browser clients and must list the origins it serves. Non-browser clients (znn-cli, SDK, Syrius over WebSocket) send no `Origin` and are unaffected.

## Tests

`node/rpc_defaults_test.go` (new):

- the default RPC config has loopback hosts, both protocols enabled, no CORS or WebSocket origins, and the `localhost` virtual host;
- the RPC stack started from the defaults (ephemeral ports) binds loopback only, answers with `Host: 127.0.0.1` and `Host: localhost`, rejects an unlisted hostname, issues no `Access-Control-Allow-Origin` for a foreign origin, and accepts WebSocket upgrades without an `Origin` or from `http://localhost` while rejecting a foreign origin;
- the exposure warning is silent for the defaults and for protocols without a listener, names each non-loopback bind and wildcard origin list, and tracks the listener rather than the enable flags when the two disagree.

All fail on the unpatched defaults.

## Verification

- `GOWORK=off go vet ./node/`
- `GOWORK=off go test -race -count=3 ./node/`
- `GOWORK=off go test ./...`
- Combined tree with #87 (`fix/honor-rpc-enable-flags`) merged in both orders: vet, build and race-tested node package pass.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,8 @@ Since version `0.0.2`, `znnd` is configured with the Alphanet Genesis and defaul
 
 Use [znn-controller](https://github.com/zenon-network/znn_controller_dart) to configure your full node. For more information please consult the [Wiki](https://github.com/zenon-network/znn-wiki).
 
+By default the HTTP (`35997`) and WebSocket (`35998`) JSON-RPC servers listen on `127.0.0.1` only and grant no cross-origin browser access. To serve other hosts, set `RPC.HTTPHost` / `RPC.WSHost` in `config.json` (or pass `--http-addr` / `--ws-addr`), and add `RPC.HTTPCors` / `RPC.WSOrigins` entries for any browser origins that need access. `znnd` logs a warning at startup for each setting that reaches beyond the local machine; put a firewall or a proxy in front of a public endpoint.
+
 ## Local devnet
 
 This branch includes a dockerized five-node local devnet with three
--- a/node/defaults.go
+++ b/node/defaults.go
@@ -20,16 +20,21 @@ var DefaultNodeConfig = Config{
 
 	LogLevel: "info",
 
+	// RPC serves local clients only unless the operator opts in: the
+	// listeners bind to loopback, HTTP accepts the localhost hostname and IP
+	// literals as Host, and no cross-origin browser access is granted (the
+	// WebSocket validator still admits localhost origins and non-browser
+	// clients when the list is empty). A public endpoint needs explicit
+	// hosts, and usually origins, in config.json or on the command line.
 	RPC: RPCConfig{
 		HTTPPort:   p2p.DefaultHTTPPort,
-		HTTPHost:   "0.0.0.0",
+		HTTPHost:   "127.0.0.1",
 		EnableHTTP: true,
 		WSPort:     p2p.DefaultWSPort,
-		WSHost:     "0.0.0.0",
+		WSHost:     "127.0.0.1",
 		EnableWS:   true,
 
-		HTTPCors:  []string{"*"},
-		WSOrigins: []string{"*"},
+		HTTPVirtualHosts: []string{"localhost"},
 	},
 	Net: NetConfig{
 		ListenHost:        p2p.DefaultListenHost,
--- a/node/node.go
+++ b/node/node.go
@@ -154,6 +154,7 @@ func (node *Node) Start() error {
 		log.Error("failed to start rpc", "reason", err)
 		return err
 	}
+	node.warnRPCExposure()
 
 	return nil
 }
--- a/node/rpc_defaults_test.go
+++ b/node/rpc_defaults_test.go
@@ -1,0 +1,258 @@
+package node
+
+import (
+	"net"
+	"net/http"
+	"strings"
+	"testing"
+	"time"
+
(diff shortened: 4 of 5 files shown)


edgepillar:
Reviewed at `67fdc7f`. This is a source-based review; I have not independently executed the reported tests.

The bare-default changes materially reduce exposure, and inspecting actual listeners for startup warnings is a useful approach. Three points would improve the security boundary and its coverage:

1. **An existing CLI mapping undermines explicit RPC opt-in.** `applyFlagsToConfig` assigns the P2P/network `--host` flag to `cfg.RPC.HTTPHost`, not `cfg.Net.ListenHost`. Starting with the new defaults and explicitly configuring `--host=0.0.0.0` therefore changes HTTP RPC to a wildcard bind without an RPC-specific override. The new warning reports the result but does not prevent it. This assignment predates the PR, but it is directly relevant to the local-only default. Please fix or explicitly track it, with a regression asserting that changing the P2P host leaves RPC hosts unchanged. [Assignment](https://github.com/zenon-network/go-zenon/blob/67fdc7f1cab2db3aae80287670ca97f1205a1e06/app/config.go#L73-L75).

2. **“Grant no cross-origin browser access” overstates the WebSocket restriction.** Empty `WSOrigins` activates the existing fallback for `http://localhost` and `http://<machine-hostname>`. Those rules omit a port, so matching HTTP origins on other ports remain allowed. For example, `http://localhost:3000` is allowed, while `http://127.0.0.1:3000` and `https://localhost:3000` do not match. Clients without an Origin header are also accepted. Please describe the removal of unrestricted browser access and document these exceptions and compatibility changes. HTTP CORS restrictions should likewise not be described as authentication. [Fallback](https://github.com/zenon-network/go-zenon/blob/67fdc7f1cab2db3aae80287670ca97f1205a1e06/rpc/server/websocket.go#L82-L108), [origin matching](https://github.com/zenon-network/go-zenon/blob/67fdc7f1cab2db3aae80287670ca97f1205a1e06/rpc/server/websocket.go#L134-L159).

3. **The defaults tests currently collapse both protocols onto one listener.** Setting both ports to zero triggers same-port sharing, so the HTTP and WebSocket listener assertions inspect the same server. Please add distinct-listener coverage, shared/HTTP-only/WS-only warning cases, and an origin table covering the cases above. Actual JSON-RPC round trips should assert the decoded result rather than only HTTP status or WebSocket upgrade success. [Test setup](https://github.com/zenon-network/go-zenon/blob/67fdc7f1cab2db3aae80287670ca97f1205a1e06/node/rpc_defaults_test.go#L24-L40).

The same-port/different-host binding behavior also needs an explicit policy alongside #87. Finally, this hardens inherited defaults, not existing configurations that explicitly retain wildcard hosts or origins, including the checked-in devnet configuration.

I did not find a newly introduced production vulnerability in the new default values or warning helper itself; the CLI issue is an inherited integration defect.

openmac0815:

The shipped defaults were a real exposure: both RPC servers on `0.0.0.0`, `*` CORS and WS origins, every public namespace — plus the empty vhost list 403'd plain localhost requests. Loopback binds + `localhost` vhost + empty origin lists fix this while keeping wallets and local tools working (empty WS origin list still admits localhost origins and non-browser clients). The startup warning per non-loopback listener / wildcard origin is the right nudge for operators who deliberately expose. Release notes should call out the default change for anyone running public RPC.

PR #89, opened 2026-09-05
rpc: bound batch size, batch response size and in-flight calls

0x3639:
## Summary

The JSON-RPC server bounded the bytes of one request (5 MiB over HTTP, 15 MiB over WebSocket) but nothing past that: a batch was decoded into one message per element however many there were, every element was executed and its response retained until the whole array was written, and on a persistent connection each incoming message started another call goroutine with no cap. Compact calls therefore turned a bounded request into an unbounded number of allocations, method invocations, response bytes and goroutines.

This PR adds the bounds the upstream RPC stack this code derives from adopted later, plus a per-connection admission cap:

| Bound | Value | Behaviour |
| --- | --- | --- |
| Batch size | 1000 calls | Counted in `parseMessage` before each element is allocated. The batch is answered with a single invalid-request error (`-32600`, "batch too large", id null, as the spec prescribes for a batch rejected as a whole) and nothing executes. Persistent connections keep serving. Invalid elements count like valid ones. |
| Batch response size | 25 MB | `handleBatch` sums the encoded size of each answer's result or error. The answer crossing the budget is still delivered; every call after it gets `-32003` "batch response too large" instead of executing. |
| Running calls | 32 per connection | Accepted messages wait for a run slot inside their own goroutine, never on the connection's dispatch loop (callbacks that make reverse calls through `ClientFromContext` need that loop to deliver their replies). |
| Accepted messages | 64 per connection | Beyond it a message (or each call of a batch) is answered at once with `-32005` "too many pending requests on this connection". |

Waiting goroutines exit when the handler closes, but take a free slot in preference to that signal because the HTTP path closes the handler right after dispatch and waits for the call. The Go client refuses to send a batch over 1000 elements itself, since the server's rejection carries no IDs it could resolve.

Known clients are unaffected: Syrius and `znn_sdk_dart` do not batch node RPC requests, and 64 outstanding messages on one socket is far beyond their concurrency.

## Tests

`rpc/server/batch_test.go` (new):

- HTTP (`httptest` over `Server.ServeHTTP`): a batch of exactly 1000 calls is served in full; 1001 is answered with the single error and executes no call; 1001 compact invalid elements are rejected while 1000 are answered individually.
- WebSocket (`Server.WebsocketHandler`): the oversized batch is rejected and the connection stays usable.
- In-process: 4 MB results are cut off after the budget with the remaining elements carrying the response-too-large error and no further execution, and the same for 4 MB error payloads; the client refuses a 1001-element batch without a round trip; 33 concurrent calls whose callbacks call back into the client all complete; 69 concurrent blocking calls see exactly 5 rejected with the overload error while 32 run and the rest wait, and every accepted call completes; 40 concurrent blocking calls leave exactly 32 inside the service.

The limit tests fail on the unpatched code; the reverse-call, error-accounting and client-refusal tests were written against the first version of this change, which they failed.

## Verification

- `GOWORK=off go vet ./rpc/server/`
- `GOWORK=off go test -race -count=5 ./rpc/server/` and `-count=3 ./rpc/api/subscribe/`
- `GOWORK=off go test ./...`
- Combined tree with #85 (`fix/bound-rpc-subscriptions`, which also edits `handler.go`) merged in both orders: identical trees, vet, build and race-tested rpc packages pass.

### DIFF
--- a/rpc/server/batch_test.go
+++ b/rpc/server/batch_test.go
@@ -1,0 +1,453 @@
+package server
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"sync"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"github.com/gorilla/websocket"
+)
+
+// batchTestService counts invocations and can hand back large results or
+// block until released, so the tests can observe execution rather than
+// only responses.
+type batchTestService struct {
+	calls   int32
+	inside  int32
+	release chan struct{}
+}
+
+func (s *batchTestService) Echo(v string) string {
+	atomic.AddInt32(&s.calls, 1)
+	return v
+}
+
+func (s *batchTestService) Big(n int) string {
+	atomic.AddInt32(&s.calls, 1)
+	return strings.Repeat("x", n)
+}
+
+// Reverse calls back into the client on the same connection; the reply
+// can only be delivered if the connection's dispatch loop keeps running.
+func (s *batchTestService) Reverse(ctx context.Context) (string, error) {
+	atomic.AddInt32(&s.calls, 1)
+	client, ok := ClientFromContext(ctx)
+	if !ok {
+		return "", fmt.Errorf("no client in context")
+	}
+	<-s.release
+	var res string
+	if err := client.CallContext(ctx, &res, "peer.echo", "back"); err != nil {
+		return "", err
+	}
+	return res, nil
+}
+
+// Fail returns an error carrying n bytes of data.
+func (s *batchTestService) Fail(n int) (string, error) {
+	atomic.AddInt32(&s.calls, 1)
+	return "", &bigDataError{strings.Repeat("e", n)}
+}
+
+type bigDataError struct{ data string }
+
+func (e *bigDataError) Error() string          { return "failed with data" }
+func (e *bigDataError) ErrorData() interface{} { return e.data }
+
+type peerService struct{}
+
+func (peerService) Echo(v string) string { return v }
+
+func (s *batchTestService) Block() string {
+	atomic.AddInt32(&s.calls, 1)
+	atomic.AddInt32(&s.inside, 1)
+	defer atomic.AddInt32(&s.inside, -1)
+	<-s.release
+	return "released"
+}
+
+func newBatchTestServer(t *testing.T) (*Server, *batchTestService) {
+	t.Helper()
+	svc := &batchTestService{release: make(chan struct{})}
+	server := NewServer()
+	if err := server.RegisterName("test", svc); err != nil {
+		t.Fatal(err)
+	}
+	t.Cleanup(server.Stop)
+	return server, svc
+}
+
+// batchOf builds a JSON array of n test.echo calls.
+func batchOf(n int) []byte {
+	var b bytes.Buffer
+	b.WriteByte('[')
+	for i := 0; i < n; i++ {
+		if i > 0 {
+			b.WriteByte(',')
+		}
+		fmt.Fprintf(&b, `{"jsonrpc":"2.0","id":%d,"method":"test.echo","params":["a"]}`, i+1)
+	}
+	b.WriteByte(']')
+	return b.Bytes()
+}
+
+// isBatchTooLarge reports whether body is the single invalid-request error
+// the server sends for an oversized batch.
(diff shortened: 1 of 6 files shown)


edgepillar:
Reviewed at `8496055`. This is a source-based review; I have not independently executed the tests or the reported combined-tree validation with #85.

The bounded admission and batch accounting are useful hardening. I found three failure-path concerns:

1. **P2: overload rejection can leave a subscription active after the Go client completes `Unsubscribe()`.**

   With 64 accepted unfinished messages, the new admission check also rejects `ledger.unsubscribe` with `-32005`, before server-side cleanup runs. The bundled Go client's forwarding loop calls `requestUnsubscribe()` but discards its error. Local unsubscribe then completes through `sync.Once`, so calling `Unsubscribe()` again does not retry remote cleanup.

   The subscription remains in the server map with its error channel open after the workload subsides. With #85 integrated, it also retains subscription capacity; a sweep of closed subscriptions cannot reclaim an entry that never became closed. This is a new overload interaction with pre-existing client error suppression, not an assertion that all calls must bypass admission limits. [Rejection](https://github.com/zenon-network/go-zenon/blob/84960552f4516ccb8251b1b69e33ddf32d111f90/rpc/server/handler.go#L164-L179), [ignored cleanup error](https://github.com/zenon-network/go-zenon/blob/84960552f4516ccb8251b1b69e33ddf32d111f90/rpc/server/subscription.go#L291-L303).

   Suggested regression: establish a subscription, use barriers to hold 32 executing and 32 queued calls, then invoke the real client `Unsubscribe()`. Verify remote removal and error-channel closure, including quota reclamation with #85. A bounded teardown path or explicit failure/recovery handling would address this without introducing unrestricted work or retries. Connection closure or a successful explicit protocol unsubscribe remains a recovery path.

2. **Overload output can block the dispatcher needed by reverse calls.**

   Accepted requests wait for execution slots off the dispatch loop, but rejected requests call `h.conn.writeJSON` synchronously from `handleMsg`/`handleBatch`. Writer backpressure or encoder-mutex contention can therefore prevent the dispatcher from delivering reverse-call responses and processing timeout notifications. The current reverse-call test uses 33 requests and never exercises the 64-message rejection boundary. [Overload writes](https://github.com/zenon-network/go-zenon/blob/84960552f4516ccb8251b1b69e33ddf32d111f90/rpc/server/handler.go#L150-L179).

   Please add a controlled-writer test that blocks overload output while an outstanding reverse response arrives. The oversized-batch rejection path also writes synchronously from the reader and deserves equivalent coverage. Keep any rejection-output mechanism bounded.

   This is a source-supported liveness risk, not a demonstrated permanent deadlock or server-wide DoS. The transport's write deadline does not bound waiting to acquire the encoder mutex.

3. **P3: invalid-request responses disappear in fallback paths.**

   For an empty batch, the return value from `startCallProc` is ignored, so `[]` becomes silent at full pending capacity. Other overload fallbacks answer only messages with valid IDs, omitting the null-ID errors normally returned for inputs such as `1`, `null` and `{}`. The response-budget fallback similarly omits invalid remaining elements. These are different from valid notifications, which should remain silent. [Empty batch](https://github.com/zenon-network/go-zenon/blob/84960552f4516ccb8251b1b69e33ddf32d111f90/rpc/server/handler.go#L103-L108), [budget and overload fallbacks](https://github.com/zenon-network/go-zenon/blob/84960552f4516ccb8251b1b69e33ddf32d111f90/rpc/server/handler.go#L132-L159).

   Please cover those invalid inputs under saturation and after response-budget exhaustion, using a valid notification as the silence control.

Two scope distinctions are worth retaining: the 25 MB value is an accumulated payload cutoff, not a hard serialized-response ceiling; and per-handler admission does not provide aggregate HTTP/server-wide resource limits. Neither is a newly discovered bypass of the stated design.

The unsubscribe interaction is my strongest merge-time concern. The existing healthy-reader tests and reported clean combination with #85 do not exercise that lifecycle boundary.

openmac0815:

All four bounds verified in the diff: batch size counted in `parseMessage` before element allocation (oversized batch costs nothing per extra element and is answered with the single `-32600`), the 25 MB response budget delivers the crossing answer and errors the rest with `-32003`, and the 32-running/64-pending per-connection caps with slot acquisition on the call goroutine — never the dispatch loop — which correctly avoids the reverse-call deadlock. The prefer-free-slot-over-close-signal detail is right for the HTTP path that closes the handler after dispatch. Client-side refusal of oversized batches avoids unresolvable replies. Merge-verified against #85 per the PR description.

PR #90, opened 2026-09-05
node: keep the producer password out of printed and logged config

0x3639:
## Summary

`MakeConfig` prints the effective configuration to standard output as JSON and hands the same value to the info log. `ProducerConfig.Password`, the password used to unlock the producer key file, was a plain exported string, so the JSON print carried it verbatim into whatever collects the node's output (terminal, service manager, container logs, support bundles). The structured log line did not carry it, but only because the configuration is logged as a value whose `Producer` field is a pointer, which `fmt` renders as an address; that is an accident of the field layout, not a guarantee.

The password is now a `node.Secret`, a string type that renders as a fixed marker under every `fmt` verb and under JSON encoding while still reading from `config.json` as a plain string. The one place that needs the value, `parseProducer`, converts it explicitly. This protects both existing sinks and any later rendering of the configuration or of the producer section, instead of patching two call sites.

Details:

- `Stringer` and `GoStringer` alone are not enough: `fmt` consults them only for the string-like verbs, and while it prints the diagnostic for a verb that is invalid for the operand it disables method lookup entirely. `Secret` therefore implements `fmt.Formatter` for every verb, and `ProducerConfig` implements it too (rendering through a method-less alias type so its layout is unchanged), so that a pointer to it nested in another value is rendered by the struct itself rather than by the diagnostic path.
- The one verb no type can intercept is `%p` applied to a non-pointer value, which `fmt` handles before consulting any method. It is documented on the type; pointer forms print an address and nothing in the repository uses the invalid form.
- The printed configuration still shows the producer section with `Password: "<redacted>"`; an empty password stays empty. The JSON key and `config.json` format are unchanged, and the whole repository builds unchanged, so no Go code assigned or read the field as a string.

**Operators:** treat existing startup output and log archives as containing the password. Restrict or remove retained copies and rotate the producer wallet password.

## Tests

- `node/secret_test.go` (new): 18 `fmt` verbs applied to the field, the producer section, the configuration, pointers to each, and wrappers, slices and maps holding them, all hide the password while non-secret fields remain; `json.Marshal` and `MarshalIndent` hide it and keep the other fields; the value round-trips from JSON and an explicit conversion yields it; an empty password renders empty.
- `app/config_test.go` (new): `MakeConfig` driven through a real `cli.Context` with a config file that sets a producer password, standard output captured and the log directory under a temporary data path read back; the configuration is printed and logged, neither sink contains the password, the producer section is still printed, and the returned configuration still carries the value.

On the unpatched code the node tests fail for the producer section, the field and both JSON encoders, and the app test fails on standard output. With `Stringer` alone, 77 subject/verb combinations still fail; the `Formatter` implementations close them.

## Verification

- `GOWORK=off go vet ./node/ ./app/`
- `GOWORK=off go build ./...`
- `GOWORK=off go test -race -count=3 ./node/ ./app/`
- `GOWORK=off go test ./...`

### DIFF
--- a/app/config_test.go
+++ b/app/config_test.go
@@ -1,0 +1,116 @@
+package app
+
+import (
+	"encoding/json"
+	"flag"
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/urfave/cli/v2"
+)
+
+const sentinel = "sentinel-producer-password-7f3a"
+
+// MakeConfig prints the effective configuration to stdout and logs it; a
+// configured producer password must reach neither sink, while the rest of
+// the configuration still does.
+func TestMakeConfigDoesNotRevealProducerPassword(t *testing.T) {
+	dataPath := t.TempDir()
+	cfgPath := filepath.Join(dataPath, "config.json")
+	file := map[string]interface{}{
+		"DataPath": dataPath,
+		"Producer": map[string]interface{}{
+			"Address":     "z1qqjnwjjpnue8xmmpanz6csze6tcmtzzdtfsww7",
+			"Index":       0,
+			"KeyFilePath": "producer",
+			"Password":    sentinel,
+		},
+	}
+	raw, err := json.Marshal(file)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if err := os.WriteFile(cfgPath, raw, 0o600); err != nil {
+		t.Fatal(err)
+	}
+
+	set := flag.NewFlagSet("znnd", flag.ContinueOnError)
+	ConfigFileFlag.Apply(set)
+	if err := set.Parse([]string{"--" + ConfigFileFlag.Name, cfgPath}); err != nil {
+		t.Fatal(err)
+	}
+	ctx := cli.NewContext(cli.NewApp(), set, nil)
+
+	stdout := captureStdout(t)
+	cfg, err := MakeConfig(ctx)
+	out := stdout()
+	if err != nil {
+		t.Fatalf("MakeConfig: %v", err)
+	}
+
+	if !strings.Contains(out, "Using the following znnd config") {
+		t.Fatalf("configuration was not printed: %q", out)
+	}
+	if strings.Contains(out, sentinel) {
+		t.Fatalf("stdout reveals the producer password:\n%s", out)
+	}
+	if !strings.Contains(out, `"KeyFilePath"`) {
+		t.Fatalf("stdout lost the producer section:\n%s", out)
+	}
+
+	logs := readLogs(t, dataPath)
+	if !strings.Contains(logs, "using znnd config") {
+		t.Fatalf("configuration was not logged:\n%s", logs)
+	}
+	if strings.Contains(logs, sentinel) {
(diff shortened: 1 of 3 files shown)


edgepillar:
Reviewed at `7425844`. This is a source-based review; I have not independently executed the reported tests.

The startup-output fix looks sound: JSON output reaches the redacting marshaler, formatting is protected at both the secret and producer-configuration levels, and `parseProducer` explicitly extracts the original value for unlock. I did not find a newly introduced password leak in those paths.

A few compatibility and test details would benefit from clarification:

1. **Distinguish JSON input compatibility from Go API and serialization compatibility.** Changing the exported `ProducerConfig.Password` field from `string` to `Secret` requires explicit conversions in downstream code assigning string variables. Also, marshaling a populated `node.Config` and unmarshaling that output yields the literal redaction marker as the password. That is appropriate for diagnostic output, but consumers should not treat the serialized configuration as a credential-preserving clone or persistence format. I did not identify an affected in-repository persistence path. [Field and marshaler](https://github.com/zenon-network/go-zenon/blob/7425844c9e7e3fda6f945ce0fd415b2bcb1853e6/node/config.go#L64-L84).

2. **Pin the behavior of `null` input.** `Secret.UnmarshalJSON` decodes into a fresh string and assigns it, so decoding `null` into a previously populated secret clears its value. The former plain-string field retained its existing value in that case. Please add a populated-target/null regression and decide whether preserving the earlier behavior is intended. This is a compatibility edge case, not a demonstrated startup failure. [Decoder](https://github.com/zenon-network/go-zenon/blob/7425844c9e7e3fda6f945ce0fd415b2bcb1853e6/node/config.go#L69-L76).

3. **Keep the protection claim and tests precise.** The introductory “every fmt verb” wording should acknowledge the already documented non-pointer `%p` exception. Additional cases for escaped/non-ASCII passwords, invalid JSON input, and an assertion that the decoded output `Password` equals the marker would strengthen the suite. The stdout capture helper should also restore `os.Stdout` through cleanup/defer and close the pipe reader, so a panic cannot leave global test state redirected. [Formatting caveat](https://github.com/zenon-network/go-zenon/blob/7425844c9e7e3fda6f945ce0fd415b2bcb1853e6/node/config.go#L22-L48), [capture helper](https://github.com/zenon-network/go-zenon/blob/7425844c9e7e3fda6f945ce0fd415b2bcb1853e6/app/config_test.go#L78-L96).

These are compatibility/documentation and test-hardening suggestions, not a request to weaken the redaction. Please also carry the existing operator guidance about historical output and password rotation into release notes: this change cannot sanitize previously retained logs.

openmac0815:

Confirmed: `Secret` redacts through `String`/`GoString`/`Format`/`MarshalJSON` while `UnmarshalJSON` keeps config-file compatibility, the explicit `string()` conversion at the unlock call is the only read path, and `ProducerConfig.Format` closes the fmt pointer-diagnostic bypass for nested structs. The end-to-end test proves the sentinel reaches neither stdout nor the log files while the returned config still carries it — that's the right assertion level.

Two things worth documenting for operators: (1) existing startup logs and archives should be treated as password-bearing and retained producer passwords rotated; (2) the `%p`-on-non-pointer limitation is inherent to fmt, fine as documented.

PR #91, opened 2026-09-05
p2p/legacy: answer pings from the keepalive loop

0x3639:
## Summary

Every ping from a legacy peer started a goroutine to write the pong, and the read loop went straight back to reading. Writes on a connection are serialized behind one mutex, so when pings arrived faster than pongs could be written, or while the remote was slow to read, those goroutines accumulated with no bound, one per ping, each holding its pending write. Their write errors were also ignored, so a connection whose writes had started failing kept accumulating writers until the read side noticed. The wait group that counted them is never waited on.

Change (`p2p/legacy/peer.go`):

- The per-peer keepalive loop, which already writes our own pings, now writes the pongs too. The read loop hands it a reply through a one-slot channel with a non-blocking send: a pong that is already pending answers every ping that arrives before it is written, so at most one pong is in flight and one is pending per peer, and a ping costs the read loop nothing beyond discarding the payload.
- A pong write error is reported on the peer's protocol-error channel exactly as a ping write error already is, so a broken connection closes promptly.
- No ping rate policy is added: with the reply coalesced, a ping is no more expensive to receive than any other base-protocol message the read loop already discards.

Liveness for legitimate peers is unchanged: pongs were already serialized behind the same RLPx write mutex, the frame write deadline (20 s) stays below the remote read deadline (30 s), and any intervening frame satisfies the remote's read. Shutdown remains bounded by the write deadline, since `rlpx.close` takes the write mutex before closing the socket.

The libp2p backend does not use this handler; this fixes the legacy backend that runs until the transport spork activates.

## Tests

`p2p/legacy/peer_test.go` (new; the package had no tests). A transport fake feeds messages from a channel and holds, fails or, after close, rejects writes while counting concurrent writers.

- 200 pings with writes held: at most one writer inside the transport and no goroutine growth to speak of; after release at most two pongs are written and a later ping is answered again.
- A failing pong write ends the peer.
- Disconnecting with a pong stuck in the transport returns the requested reason and leaves no writer behind.

On the unpatched code the first test observes 200 concurrent writers and the second sees the peer keep running after the write failure.

## Verification

- `GOWORK=off go vet ./p2p/legacy/`
- `GOWORK=off go test -race -count=3 ./p2p/legacy/`
- `GOWORK=off go test ./...`

### DIFF
--- a/p2p/legacy/peer.go
+++ b/p2p/legacy/peer.go
@@ -68,6 +68,11 @@ type Peer struct {
 	protoErr chan error
 	closed   chan struct{}
 	disc     chan p2p.DiscReason
+
+	// pongPending holds at most one pending pong reply; pings that arrive
+	// while it is full are answered by that same pong. The keepalive loop
+	// is the only writer of pongs.
+	pongPending chan struct{}
 }
 
 // NewPeer returns a peer for testing purposes.
@@ -128,6 +133,8 @@ func newPeer(conn *conn, protocols []p2p.Protocol) *Peer {
 		disc:     make(chan p2p.DiscReason),
 		protoErr: make(chan error, len(protomap)+1), // protocols + pingLoop
 		closed:   make(chan struct{}),
+
+		pongPending: make(chan struct{}, 1),
 	}
 	return p
 }
@@ -201,6 +208,10 @@ loop:
 	return reason
 }
 
+// pingLoop sends our periodic pings and the pong replies to the remote
+// peer's pings, so base-protocol writes for a peer come from one goroutine
+// and a write that cannot complete holds up further replies instead of
+// piling up writers.
 func (p *Peer) pingLoop() {
 	ping := time.NewTicker(pingInterval)
 	defer ping.Stop()
@@ -208,6 +219,11 @@ func (p *Peer) pingLoop() {
 		select {
 		case <-ping.C:
 			if err := p2p.SendItems(p.rw, pingMsg); err != nil {
+				p.protoErr <- err
+				return
+			}
+		case <-p.pongPending:
+			if err := p2p.SendItems(p.rw, pongMsg); err != nil {
 				p.protoErr <- err
 				return
 			}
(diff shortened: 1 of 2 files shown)


edgepillar:
Reviewed at `846abc3`. This is a source-based review; I have not independently executed the reported tests.

The one-slot handoff correctly bounds pending pong work, and reporting pong write failures through the existing keepalive error path is a useful improvement. I did not find a newly introduced production correctness or security defect in the changed scheduling.

Two test improvements would make the evidence stronger:

1. **Make the coalescing test deterministic.** The sleeps and process-wide goroutine count do not establish when the pending pong has completed. In particular, an old pending pong and the later ping's reply can both complete between counter observations, producing `sent+2` and failing the `count == sent+1` expectation on otherwise correct code. Explicit writer-entry/completion and read-handling barriers would avoid that ambiguity. Please also exercise a controlled periodic-ping tick while pong work is pending; the current tests normally finish before the real 15-second ticker fires. [Coalescing test](https://github.com/zenon-network/go-zenon/blob/846abc35d74494af3446559a8b95059ddc1c67a5/p2p/legacy/peer_test.go#L145-L165).

2. **Distinguish fake-transport cleanup from RLPx shutdown timing.** `heldTransport.close` immediately releases blocked writers, whereas `rlpx.close` acquires the same write mutex as `WriteMsg` before closing the socket. The existing per-write deadline therefore does not establish a strict end-to-end shutdown deadline, including time waiting behind other writers. A controlled transport test modeling the shared lock, competing protocol output and deadline expiry would provide stronger evidence. Alternatively, narrow the test description to the eventual cleanup it actually demonstrates. [Shutdown test](https://github.com/zenon-network/go-zenon/blob/846abc35d74494af3446559a8b95059ddc1c67a5/p2p/legacy/peer_test.go#L182-L201), [RLPx locking](https://github.com/zenon-network/go-zenon/blob/846abc35d74494af3446559a8b95059ddc1c67a5/p2p/legacy/rlpx.go#L105-L130).

The locking behavior predates this PR. Likewise, `Peer.run` still does not wait for its full wait group, so the change should not be described as proving that every peer goroutine has joined when `run` returns.

For scope clarity, this bounds pong scheduling rather than incoming-frame CPU or memory. #92 complements it with an individual frame-size bound; neither change alone establishes aggregate resource limits. These are nonblocking coverage and wording suggestions.

openmac0815:

The old per-ping goroutine spawn was a genuine remote-triggerable pile-up: concurrent `WriteMsg` calls share the RLPx cipher/MAC state, so this was both a race/frame-corruption vector and an unbounded-goroutine vector behind a stalled transport. Routing pongs through the keepalive loop with the one-slot `pongPending` channel restores a single base-protocol writer, coalescing is protocol-safe (devp2p pings are liveness probes, not accounting), and pong errors now terminate the peer via `protoErr` instead of being dropped. The close-aware paths avoid the send deadlock. Tests cover held writes, ping floods, error propagation and disconnect-with-blocked-pong — deterministic, no sleeps.

No conflict with #92 (different files, complementary).

PR #92, opened 2026-09-05
p2p/legacy: bound the RLPx frame size before allocating it

0x3639:
## Summary

The RLPx frame reader authenticated and decrypted the 32-byte header, took the 24-bit content length from it and allocated a buffer of that length (rounded up to 16 bytes) before reading the body. The length is the sender's choice, so a peer that has completed the handshakes could announce the wire maximum and cause a 16 MiB allocation with a single header, then send nothing and hold it until the 30 second read deadline. The protocol layer's 10 MiB message limit runs only after the whole frame and its MAC have been read into that buffer. The libp2p backend already checks its length before allocating; the legacy backend, which runs until the transport spork activates, did not.

Change (`p2p/legacy/rlpx.go`): `ReadMsg` rejects a header whose content length exceeds `maxFrameSize` before computing the padded size or allocating anything. `maxFrameSize` is the protocol layer's 10 MiB message limit (repeated here as `maxMessageSize`, the same way `p2p/libp2p/stream_rw.go` does, since the transport cannot import the protocol package) plus the RLP-encoded message code of at most 9 bytes that shares the frame with the payload. The header MAC check still comes first, and the protocol layer's own check remains as defense in depth: it still rejects a frame that fits the transport bound but carries a payload over its limit.

Compatibility: every protocol-valid payload remains readable, including a 10 MiB payload with a maximal 9-byte code. No legitimate message class occupies the 10 to 16 MiB range; such a frame was previously read in full and then rejected by the protocol layer with a disconnect, and is now rejected by the transport with a read error and the same disconnect. The encryption handshake uses raw fixed-size I/O and is untouched; the protocol handshake uses the framed reader with its own stricter 2 KiB limit.

## Tests

`p2p/legacy/rlpx_test.go` (new): a writer and a reader share one in-memory wire and the same secrets, as the upstream fake-frame test does; the wire records whether the reader ever asked for bytes past the header.

- A header announcing one byte over the limit, and one announcing the 24-bit maximum, are each rejected with a frame size error, no body byte is requested and less than 1 MiB is allocated in the process.
- The largest message the protocol layer accepts (10 MiB payload) still passes the transport and round-trips, while one byte more is rejected without a body read.
- A header with a tampered MAC is still rejected first, and an ordinary frame round-trips.

On the unpatched code the oversized headers lead to a body read that fails on the empty wire after the allocation, and the frame one byte over the limit is accepted.

## Verification

- `GOWORK=off go vet ./p2p/legacy/` (only the pre-existing unkeyed `ecdsa.PublicKey` literal warning, present at the base)
- `GOWORK=off go test -race -count=3 ./p2p/legacy/`
- `GOWORK=off go test ./...`
- Combined tree with #91 (`fix/coalesce-peer-pongs`, same package): identical trees in both merge orders, vet, build and race-tested package pass.

### DIFF
--- a/p2p/legacy/rlpx.go
+++ b/p2p/legacy/rlpx.go
@@ -46,6 +46,14 @@ import (
 const (
 	maxUint24 = ^uint32(0) >> 8
 
+	// maxMessageSize matches protocol.ProtocolMaxMsgSize, the largest
+	// payload the protocol layer accepts, and maxFrameSize adds the
+	// RLP-encoded message code (at most 9 bytes) that shares the frame
+	// with the payload. A frame header announcing more is rejected before
+	// any buffer of the announced size is allocated.
+	maxMessageSize = 10 * 1024 * 1024
+	maxFrameSize   = maxMessageSize + 9
+
 	sskLen = 16 // ecies.MaxSharedKeyLength(pubKey) / 2
 	sigLen = 65 // elliptic S256
 	pubLen = 64 // 512 bit pubkey in uncompressed representation without format byte
@@ -585,6 +593,12 @@ func (rw *rlpxFrameRW) ReadMsg() (msg p2p.Msg, err error) {
 	rw.dec.XORKeyStream(headbuf[:16], headbuf[:16]) // first half is now decrypted
 	fsize := readInt24(headbuf)
 	// ignore protocol type for now
+
+	// The header is authenticated but its length is the sender's choice;
+	// bound it before sizing anything by it.
+	if fsize > maxFrameSize {
+		return msg, fmt.Errorf("frame size %d exceeds maximum %d", fsize, maxFrameSize)
+	}
 
 	// read the frame content
 	var rsize = fsize // frame size rounded up to 16 byte boundary
--- a/p2p/legacy/rlpx_test.go
+++ b/p2p/legacy/rlpx_test.go
@@ -1,0 +1,156 @@
+package legacy
+
+import (
+	"bytes"
+	"crypto/rand"
+	"io"
+	"runtime"
+	"strings"
+	"testing"
+
+	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/ethereum/go-ethereum/rlp"
+	"golang.org/x/crypto/sha3"
+
+	"github.com/zenon-network/go-zenon/p2p"
+)
+
+// countingConn is the wire between a frame writer and reader. It records
+// how many bytes the reader asked for after the 32-byte header, so a test
+// can tell whether the reader tried to read a frame body at all.
+type countingConn struct {
+	bytes.Buffer
+	read      int
+	bodyReads int
+}
+
+func (c *countingConn) Read(p []byte) (int, error) {
+	if c.read >= 32 {
+		c.bodyReads++
+	}
+	n, err := c.Buffer.Read(p)
+	c.read += n
+	return n, err
+}
+
+func (c *countingConn) bodyRequested() bool { return c.bodyReads > 0 }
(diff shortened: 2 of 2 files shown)


edgepillar:
Reviewed at `42cbf9f`. This is a source-based review; I have not independently executed the tests or the reported combined-tree validation with #91.

The production check looks correct: header authentication precedes the size check, the bound is enforced before padding and allocation, and the nine-byte allowance preserves a 10 MiB payload alongside the largest encoded message code. I found no newly introduced cryptographic regression or rejection of an otherwise valid payload.

I would tighten the following regression evidence:

1. **Separate the transport boundary from the protocol payload boundary.** `TestReadMsgFrameLimitMatchesLargestMessage` uses code `0x10`, whose RLP encoding occupies one byte. Consequently, `largest = maxFrameSize - len(ptype)` is **10 MiB + 8 bytes of payload**, not the protocol's 10 MiB maximum. Its one-byte-over case contains 10 MiB + 9 payload bytes. This is a useful transport-boundary test, but its name/comment and the PR's test description conflate the two limits. [Boundary test](https://github.com/zenon-network/go-zenon/blob/42cbf9fec1d363a642f255e0958ae58678bf795c/p2p/legacy/rlpx_test.go#L93-L126).

   Please add explicit cases for a short code with exactly 10 MiB and a nine-byte code with exactly 10 MiB. Also distinguish the one-byte-over cases: a short-code payload of 10 MiB + 1 still fits the frame allowance and must be rejected by protocol validation; the same payload with a nine-byte code should be rejected before reading the body.

2. **Test authentication precedence using an oversized header.** The current tampered-header test announces only 64 bytes. A header announcing more than `maxFrameSize` with a corrupted MAC would additionally verify that authentication takes precedence over the new size check, without requesting any body bytes. [MAC test](https://github.com/zenon-network/go-zenon/blob/42cbf9fec1d363a642f255e0958ae58678bf795c/p2p/legacy/rlpx_test.go#L132-L141).

3. **Retain the direct behavioral allocation guard.** The “no body requested” assertion is valuable. Process-wide `runtime.TotalAlloc` can include unrelated background allocation, so the allocation threshold should not be the sole regression signal. Truncated frames, corrupted body MACs and consecutive valid frames would also strengthen the in-memory fixture's coverage.

There is one pre-existing follow-up opportunity: `readProtocolHandshake` calls framed `ReadMsg` before applying its 2 KiB handshake limit. A peer that has completed encryption can therefore still trigger an approximately 10 MiB allocation before completing protocol admission, subject to the existing handshake deadline. This PR reduces the previous maximum but does not eliminate that gap. A handshake-specific preallocation limit, tested with an authenticated header and no body, would address it while preserving the ordinary post-handshake allowance. [Handshake ordering](https://github.com/zenon-network/go-zenon/blob/42cbf9fec1d363a642f255e0958ae58678bf795c/p2p/legacy/rlpx.go#L154-L161).

That remaining handshake exposure is not introduced by this patch and should be tracked separately from the correctness of the new steady-state frame bound.

openmac0815:

Placement verified: the 24-bit frame length is checked after header MAC authentication (forged headers rejected first) and before any allocation or body read. The bound (`10 MiB + 9` for the RLP message-code prefix) matches the protocol maximum, and the boundary tests pin limit-exactly vs limit+1. Rejecting before consuming the body can't desynchronize a surviving connection because the read error terminates the peer. This removes the invalid ~10–16 MiB allocation window for authenticated peers; the protocol-layer limit remains as defense in depth.

Optional follow-up: a symmetric check on the direct write path, should anything ever write raw frames outside the protocol send paths.

PR #93, opened 2026-09-05
rpc/embedded: enrich only the requested page of projects and requests

0x3639:
## Summary

Two public embedded RPC listings did per-entry work for the whole collection and only then sliced the requested page.

- **`embedded.accelerator.getAll`** listed every project, sorted them, then ran the vote breakdown scan and the phase lookups for every project before slicing. The listing and the sort still need every project (the order is by last update, which is not the storage order), but the lookups are now done only for the projects on the page, so an empty or one-item page costs the listing and nothing more. Page contents, order and `count` are unchanged.
- **`embedded.bridge.getAllUnsignedWrapTokenRequests`** built the enriched entry (token lookup and confirmations to finality) for every unsigned request before reversing and slicing. The selection (unsigned only, newest first, the requested page, the total count) is now a separate function run first, and the lookups are done for the page only. The other listing methods already enriched only their page.

One output difference: an empty unsigned-request page now serializes as `"list": []` rather than `"list": null`, because the page slice is always allocated. The pinned Dart SDK model and the Go consumers accept both. Non-empty output is byte-for-byte unchanged.

## Scope

The remaining full-collection cost in these methods is the storage scan that lists the collection, which the sort and the total count require. Moving that into indexed or paginated storage would change how the contracts lay out their state and is out of scope.

The caller-sized hash list of `getPillarVotes` is deliberately left as it is. Syrius sends every phase ID of a full page of projects in one call and nothing on-chain caps phases per project, so any finite server bound has to follow a client-side batching change in the SDK and Syrius. That is tracked as a follow-up.

## Tests

`rpc/api/embedded/collection_work_test.go` (new): on a mock node with the accelerator spork active and six projects created, `getAll` is driven through a context whose storage counts point lookups and prefix scans. The cost grows from an empty page to a one-item page to the full page, the one-item page costs at most one project's lookups beyond the listing, and a later page still returns the right slice with votes. The unsigned wrap selection is checked for order, count, paging past the end and the full page.

On the unpatched code the `getAll` cost is the same for every page size (one scan per project plus the listing).

## Verification

- `GOWORK=off go vet ./rpc/api/embedded/`
- `GOWORK=off go test -race -count=3 ./rpc/api/embedded/`
- `GOWORK=off go test ./vm/embedded/tests/ ./rpc/...` (golden embedded API suite)
- `GOWORK=off go test ./...`

### DIFF
--- a/rpc/api/embedded/accelerator.go
+++ b/rpc/api/embedded/accelerator.go
@@ -171,7 +171,10 @@ func (a *AcceleratorApi) GetAll(pageIndex, pageSize uint32) (*ProjectList, error
 	if err != nil {
 		return nil, err
 	}
+	return a.getAll(context, pageIndex, pageSize)
+}
 
+func (a *AcceleratorApi) getAll(context vm_context.AccountVmContext, pageIndex, pageSize uint32) (*ProjectList, error) {
 	projects, err := definition.GetProjectList(context.Storage())
 	if err != nil {
 		return nil, err
@@ -181,17 +184,16 @@ func (a *AcceleratorApi) GetAll(pageIndex, pageSize uint32) (*ProjectList, error
 		return projects[i].LastUpdateTimestamp > projects[j].LastUpdateTimestamp
 	})
 
+	// The listing and the sort need every project; the vote and phase
+	// lookups are only done for the page that is returned.
+	start, end := api.GetRange(pageIndex, pageSize, uint32(len(projects)))
 	result := &ProjectList{
 		Count: len(projects),
-		List:  make([]*Project, len(projects)),
+		List:  make([]*Project, 0, end-start),
 	}
-
-	for index, project := range projects {
-		result.List[index] = a.toProject(context, project)
+	for _, project := range projects[start:end] {
+		result.List = append(result.List, a.toProject(context, project))
 	}
-
-	start, end := api.GetRange(pageIndex, pageSize, uint32(len(result.List)))
-	result.List = result.List[start:end]
 
 	return result, nil
 }
@@ -233,6 +235,7 @@ func (a *AcceleratorApi) GetVoteBreakdown(id types.Hash) (*definition.VoteBreakd
 	}
 	return voteBreakdown, nil
 }
+
 func (a *AcceleratorApi) GetPillarVotes(name string, hashes []types.Hash) ([]*definition.PillarVote, error) {
 	_, context, err := api.GetFrontierContext(a.chain, types.AcceleratorContract)
 	if err != nil {
--- a/rpc/api/embedded/bridge.go
+++ b/rpc/api/embedded/bridge.go
@@ -455,7 +455,6 @@ func (a *BridgeApi) GetAllUnsignedWrapTokenRequests(pageIndex, pageSize uint32) 
 	if err != nil {
 		return nil, err
 	}
-	var unsignedRequests []*WrapTokenRequest
 
 	momentum, err := context.GetFrontierMomentum()
 	if err != nil {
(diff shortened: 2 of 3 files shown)


edgepillar:
Reviewed at `a8cae05`. This is a source-based review; I have not independently executed the reported tests.

Moving enrichment after page selection looks correct. The public page-size checks remain in place, the accelerator still uses the same stable sort, and the bridge selection preserves the previous filter/reverse/slice behavior. I did not find a newly introduced production correctness or consensus defect in these changes.

I would address the following points:

1. **Restore the spork globals modified by the new test.** `TestAcceleratorGetAllEnrichesOnlyThePage` writes `ImplementedSporksMap`, and `activateAcceleratorSpork` changes both that map and `AcceleratorSpork.SporkId`. `StopPanic` does not restore these values. Please snapshot and restore them through `t.Cleanup`, registered before the first mutation at line 104. Otherwise, subsequent tests or repetitions in the same process inherit modified activation state. The existing `saveSporkState` helper in the embedded test suite provides a useful pattern. This is a test-isolation issue, not a production consensus change. [New mutations](https://github.com/zenon-network/go-zenon/blob/a8cae05a4145d6d57fa963d5c02b1df04588f6c8/rpc/api/embedded/collection_work_test.go#L79-L107), [existing cleanup pattern](https://github.com/zenon-network/go-zenon/blob/a8cae05a4145d6d57fa963d5c02b1df04588f6c8/vm/embedded/tests/helpers_test.go#L12-L29).

2. **Check the ordering description against actual storage order.** Wrap-request keys encode `MaxInt64 - CreationMomentumHeight`, and `GetWrapTokenRequests` iterates them forwards. For ordinary heights such as 10, 20 and 30, storage therefore returns 30, 20, 10; the helper reverses this to 10, 20, 30. That preserves the old unsigned-list behavior, but does not establish the newly documented “newest first” claim. The fixture constructs requests in increasing ID order without setting their creation heights, so it cannot verify that claim. Please add a storage-backed ordering fixture and correct the description; I would avoid changing established ordering inside this performance-only patch. [Storage keys and iteration](https://github.com/zenon-network/go-zenon/blob/a8cae05a4145d6d57fa963d5c02b1df04588f6c8/vm/embedded/definition/bridge.go#L717-L768), [selection fixture](https://github.com/zenon-network/go-zenon/blob/a8cae05a4145d6d57fa963d5c02b1df04588f6c8/rpc/api/embedded/collection_work_test.go#L146-L170).

3. **Extend the regression evidence to the actual enrichment paths.** The bridge test exercises selection only: it would not detect enrichment being moved back before pagination in the public method. Please assert that the public call performs token enrichment only for selected requests, with zero such lookups for an empty page. For the accelerator, projects containing phases and actual votes would strengthen the current fixture, which creates projects without phases and checks only a non-nil vote breakdown. Exact returned IDs/order and vote values would establish compatibility more directly. Please also pin the `[]` JSON result when there are no unsigned requests.

4. **Consider avoiding the additional full-collection pointer allocation.** `selectUnsignedWrapRequests` reserves capacity for every stored request, even when all are signed or the requested page is empty. A count-and-page selection could retain the necessary full scan while allocating only for the returned page. This is a nonblocking memory optimization, not evidence that the overall enrichment improvement is ineffective. [Selection allocation](https://github.com/zenon-network/go-zenon/blob/a8cae05a4145d6d57fa963d5c02b1df04588f6c8/rpc/api/embedded/bridge.go#L488-L496).

The scope distinction in the PR is important: whole-collection scans and sorting remain, and a project can contain many phases. This reduces avoidable work; it does not establish a fixed per-call CPU or memory bound.

openmac0815:

Verified the semantics are preserved: accelerator keeps the full listing + stable sort with `GetRange` before `toProject`, so `Count` and page contents are unchanged while vote/phase lookups drop to page size. The bridge helper reproduces the previous newest-first order and total unsigned count, and no longer mutates the storage-returned slice. One deliberate behavior change to be aware of: off-page token-lookup failures no longer fail a page that doesn't contain the bad request — a defensible robustness improvement, but callers relying on global error behavior should note it.

Minor test-hygiene follow-up: the accelerator test mutates package-global spork state without restoring it.

PR #94, opened 2026-09-05
app: map the --host flag to the P2P listen host

0x3639:
## Summary

The `--host` flag is documented as "Network listening host", sits next to `--port` and defaults to the P2P listen host, but `applyFlagsToConfig` wrote its value into `RPC.HTTPHost`. Nothing on the command line ever set `Net.ListenHost`, so the flag could not confine the P2P listener, and an explicit `--host` silently re-bound HTTP-RPC to the address given for P2P, overriding whatever `HTTPHost` the configuration file specified. `--port` already reached `Net.ListenPort`, and `--http-addr` / `--ws-addr` already reached the RPC hosts. The mapping dates from the initial commit.

The flag now writes `Net.ListenHost`, which flows through `makeNetConfig` and the switcher into both the legacy and the libp2p listener. Anyone who had discovered the old behaviour and used `--host` to set the HTTP-RPC address should use `--http-addr`, which has always been the documented flag. No deployment command in the repository passes `--host`, and znn-controller starts a bare `znnd`.

## Tests

`app/flags_test.go` (new): each of `--host`, `--port`, `--http-addr` and `--ws-addr` is applied on its own to a default configuration through a real `cli.Context` and must change exactly the field its name describes; with no flags passed, all three hosts keep their defaults, so the flags' own default values do not leak into the configuration. On the unpatched code `--host` leaves `Net.ListenHost` untouched and changes `RPC.HTTPHost`.

## Verification

- `GOWORK=off go vet ./app/`
- `GOWORK=off go test -race -count=3 ./app/`
- `GOWORK=off go test ./...`
- Combined tree with #90 (`fix/redact-producer-password`, which also adds a test file to `app/`): identical trees in both merge orders, app and node packages race-tested.

### DIFF
--- a/app/config.go
+++ b/app/config.go
@@ -71,7 +71,7 @@ func applyFlagsToConfig(ctx *cli.Context, cfg *node.Config) {
 	}
 
 	if listenHost := ctx.String(ListenHostFlag.Name); ctx.IsSet(ListenHostFlag.Name) && len(listenHost) > 0 {
-		cfg.RPC.HTTPHost = listenHost
+		cfg.Net.ListenHost = listenHost
 	}
 
 	if ctx.IsSet(ListenPortFlag.Name) {
--- a/app/flags_test.go
+++ b/app/flags_test.go
@@ -1,0 +1,65 @@
+package app
+
+import (
+	"flag"
+	"testing"
+
+	"github.com/urfave/cli/v2"
+
+	"github.com/zenon-network/go-zenon/node"
+)
+
+// applyFlags parses args against the given flags and applies them to a
+// default configuration, the way MakeConfig does after reading the file.
+func applyFlags(t *testing.T, flags []cli.Flag, args ...string) node.Config {
+	t.Helper()
+	set := flag.NewFlagSet("znnd", flag.ContinueOnError)
+	for _, f := range flags {
+		if err := f.Apply(set); err != nil {
+			t.Fatal(err)
+		}
+	}
+	if err := set.Parse(args); err != nil {
+		t.Fatal(err)
+	}
+	cfg := node.DefaultNodeConfig
+	applyFlagsToConfig(cli.NewContext(cli.NewApp(), set, nil), &cfg)
+	return cfg
+}
+
+// Each listen flag sets the field its name and usage describe and nothing
+// else: the network flags the P2P listener, the RPC flags the RPC hosts.
+func TestListenFlagsMapToTheirFields(t *testing.T) {
+	flags := []cli.Flag{ListenHostFlag, ListenPortFlag, RPCListenAddrFlag, RPCPortFlag, WSListenAddrFlag, WSPortFlag}
+	def := node.DefaultNodeConfig
+
+	cfg := applyFlags(t, flags, "--"+ListenHostFlag.Name, "10.0.0.5")
+	if cfg.Net.ListenHost != "10.0.0.5" {
+		t.Errorf("--%s left Net.ListenHost at %q", ListenHostFlag.Name, cfg.Net.ListenHost)
+	}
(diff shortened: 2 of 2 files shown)


edgepillar:
Reviewed at `3f2af77`. This is a source-based review; I have not independently executed the reported tests or combined-tree validation with #90.

The assignment now matches the documented purpose of `--host`. I traced it through `Net.ListenHost`, `makeNetConfig` and the switcher to both P2P backends. The `IsSet` guard remains intact, and the HTTP and WebSocket address flags remain independent. I did not find a correctness defect in the one-line mapping correction.

Three details deserve attention:

1. **Make the HTTP-RPC migration consequence explicit in durable upgrade guidance.** With otherwise-default configuration, an existing `--host 127.0.0.1` command previously restricted HTTP-RPC. After this change it restricts P2P instead, while HTTP-RPC remains enabled with its default `0.0.0.0` bind. Operators relying on the old behavior need an explicit `--http-addr 127.0.0.1` or equivalent RPC configuration. The PR description already mentions the replacement flag; a concrete before/after example in release or deployment documentation would help prevent an unintended exposure change. This is a migration risk from correcting the semantics, not a newly introduced default. [Mapping](https://github.com/zenon-network/go-zenon/blob/3f2af77fb61e2a8b4c115ce8b94d4a8cf4da83f4/app/config.go#L73-L88), [RPC defaults](https://github.com/zenon-network/go-zenon/blob/3f2af77fb61e2a8b4c115ce8b94d4a8cf4da83f4/node/defaults.go#L23-L32).

2. **Test preservation of non-default configuration values.** The no-flags test starts from `DefaultNodeConfig`, whose P2P host equals the flag's own default. It therefore cannot detect the flag default accidentally overwriting a host loaded from a configuration file. Please let the helper accept an initial configuration containing distinct, non-default P2P, HTTP and WebSocket hosts and ports, and verify that omitted flags preserve them. A combined host/address/port case would also establish independent explicit overrides. These can remain pure configuration tests without starting listeners or writing files. [Test helper](https://github.com/zenon-network/go-zenon/blob/3f2af77fb61e2a8b4c115ce8b94d4a8cf4da83f4/app/flags_test.go#L14-L28), [no-flags assertion](https://github.com/zenon-network/go-zenon/blob/3f2af77fb61e2a8b4c115ce8b94d4a8cf4da83f4/app/flags_test.go#L59-L64).

3. **Clarify the supported address forms across both backends.** The downstream address builder concatenates host and port rather than using `net.JoinHostPort`. Raw IPv6 therefore requires care. There is also a backend difference: legacy can resolve `localhost`, whereas libp2p explicitly rejects DNS hostnames. A hostname may consequently work before activation and fail when the backend switches. These limitations already exist for `Net.ListenHost`; the corrected flag now reaches that path. Documenting accepted forms and adding focused address-construction/validation cases would help, without requiring DNS support or a broader networking redesign. [Address construction](https://github.com/zenon-network/go-zenon/blob/3f2af77fb61e2a8b4c115ce8b94d4a8cf4da83f4/node/node.go#L101-L108), [libp2p address contract](https://github.com/zenon-network/go-zenon/blob/3f2af77fb61e2a8b4c115ce8b94d4a8cf4da83f4/p2p/libp2p/server.go#L1519-L1537).

I support the mapping correction. My priorities would be the operator migration note and the non-default configuration regression cases.

openmac0815:

Confirmed: `--host` is documented as the network listen host but was wired to `RPC.HTTPHost` — the P2P listener ignored it while the HTTP-RPC bind was silently overridden. The one-line fix maps it to `Net.ListenHost`, and the regression test exercises the real CLI flag set with cross-field isolation and the no-flags default case.

This is an intentional compatibility break for anyone who used the broken flag to configure HTTP-RPC; the release notes should point them at the RPC listen flag.

PR #95, opened 2026-09-05
fetcher: keep the per-peer announce counter exact across a fetch

0x3639:
## Summary

Follow-up to #84, where the fetcher's announce counter drift was noted and deferred, and which a reviewer asked to have tracked separately.

The fetcher caps the announces it retains for one peer at `hashLimit` (256) using a per-peer counter. When an announce timer expires, the loop calls `forgetHash`, which decrements the counter for every announcer of that hash, and then moves the chosen announce into the `fetching` table without counting it again. When the fetch later completes or times out, `forgetHash` decrements the counter once more for the fetching entry. Every announced hash that reaches the fetching table therefore ends its lifecycle one below where it started. Because a counter was only deleted when it reached exactly zero, one that passed through zero stayed negative and kept falling, so each completed batch widened the set of announces the peer could hold next time, without bound.

## Changes

- **Transition into `fetching` counts the entry.** The chosen peer's counter is incremented when its announce becomes a fetch, since `forgetHash` released the announce and will release the fetching entry on completion or timeout. Announced plus fetching entries per peer now equal the counter at every point, so a peer with `hashLimit` hashes in flight cannot announce more until some complete.
- **`forgetHash` floors at zero.** A counter at or below zero is deleted, so no future miscount can leave a peer with a negative balance.
- **Expired fetches are swept on a timer.** The sweep ran only at the top of the loop, so with nothing else waking the loop after `fetchTimeout`, stale entries stayed counted until the next event. Since fetching entries are now charged to their peer, that next event could be the peer's own announce, judged against a count that still included timed-out fetches: a peer whose whole batch expired while the loop was idle had its next announce rejected. The sweep now records the oldest surviving fetch as it walks the table and arms a timer for it, so the loop wakes when that fetch times out. The table is walked once per iteration, as it was before the timer existed. Two alternatives were considered: sweeping inside the announce handler on every announce would make each message, including refused ones, cost a walk of the fetching table; sweeping only when an announce is about to be refused would fix the rejection but would leave timed-out entries in memory until some event arrived. The timer releases them on time either way.
- **Admission releases timed-out fetches first.** The sweep runs before the loop blocks in `select`. If the oldest fetch times out while the loop is blocked and a notification becomes ready alongside the expiry timer, `select` can take the notification first, and a peer at its limit would be refused against fetches that have already timed out. The notify path now calls `releaseExpired` for the announcing peer: when that peer is at its allowance and the oldest pending fetch is past its timeout, the sweep runs before counting. The sweep records the oldest surviving fetch's announce time, so the check costs nothing when nothing is due, and since the sweep only runs from this path when something has actually expired, a peer at its limit cannot make refused announces cost repeated walks of the table.
- **`hashLimit` becomes `HashLimit`** with the same wording as #84, so the two branches merge cleanly in either order and the tests here compile against both. Verified by applying both onto `dev` and running the protocol tests under `-race`.

go-ethereum's block fetcher has the same double decrement and only applies the zero floor. That stops the counter going negative but still lets a peer with `hashLimit` hashes in flight announce `hashLimit` more; the increment here makes the counter cover announced plus fetching, which is what the limit's comment says it bounds.

No wire format or peer-visible behaviour changes except that a peer can no longer exceed its announce allowance.

## Tests

`protocol/fetcher/fetcher_test.go` drives whole lifecycles through the real loop and, once the loop is stopped, compares every counter with the announced and fetching entries attributed to that peer, failing on any negative or stale value. The tests wait on the fetcher's `fetchingHook`, and on a new `expiredHook` that the sweep calls after removing timed-out entries, rather than sleeping past `arriveTimeout` or `fetchTimeout`. A slow loop delays them instead of failing them, and the two timeout tests observe the sweep rather than infer it from the clock. Both hooks are nil outside tests. The tests:

- Announce, fetch, and delivery of an already imported block leaves no counter behind. On the unpatched code the counter reads -8 for eight hashes.
- Announce, fetch, and `fetchTimeout` leaves no counter behind, with nothing but the timer waking the loop. Skipped under `-short` since it waits out the timeout.
- With every fetch expired while the loop is idle, all of the peer's next announces are accepted, checked as the exact set held across announced and fetching so a timer firing meanwhile cannot fail it. Also skipped under `-short`. Before the timer, the first was rejected (7 of 8 accepted); without the timer, both this and the previous test now fail at the expiry deadline with no hashes swept.
- Two direct tests on a fetcher with no loop build the state between sweep and `select` by hand: with `HashLimit` fetches from one peer all past `fetchTimeout`, `releaseExpired` frees them and the peer's next announce would be accepted; with live fetches it changes nothing and the peer stays refused, and a peer under its limit triggers no sweep even with stale fetches. These do not depend on `select` ordering or on waiting for the expiry hook. The one-line call site in the notify case is not itself pinned by a test, since forcing a notification to win the `select` against a ready timer would need control over `select` ordering or a clock abstraction.
- With `hashLimit` hashes fetching, further announces from the same peer are refused. This test fails if only the zero floor is applied, so the exact accounting is what is pinned.
- After two completed batches of `hashLimit`, the peer still holds at most `hashLimit` announces. On the unpatched code 264 are accepted.

The five lifecycle tests fail on the unpatched code. The two timeout tests also fail with the accounting fix applied but the timer removed, and the first direct test fails without the admission-time release.

## Verification

`GOWORK=off`: gofmt and vet clean; `go test -race -count=3 ./protocol/...`; `golangci-lint run ./protocol/fetcher/...` reports the same issues before and after, none on changed lines.

### DIFF
--- a/protocol/fetcher/fetcher.go
+++ b/protocol/fetcher/fetcher.go
@@ -231,13 +231,18 @@ func (f *Fetcher) Filter(peer string, blocks []*nom.DetailedMomentum) []*nom.Det
 func (f *Fetcher) loop() {
 	// Iterate the block fetching until a quit is requested
 	fetch := time.NewTimer(0)
+	expire := time.NewTimer(0)
 	for {
-		// Clean up any expired block fetches
+		// Clean up any expired block fetches. Expired entries still count
+		// against their peer's announce allowance, so the sweep must also
+		// run when nothing else wakes the loop: arm a timer for the oldest
+		// fetch so the allowance is released on time.
 		for hash, announce := range f.fetching {
 			if time.Since(announce.time) > fetchTimeout {
 				f.forgetHash(hash)
 			}
 		}
+		f.rescheduleExpiry(expire)
 		// Import any queued blocks that could potentially fit
 		height := f.chainHeight()
 		for !f.queue.Empty() {
@@ -335,6 +340,10 @@ func (f *Fetcher) loop() {
 			// Schedule the next fetch if blocks are still pending
 			f.reschedule(fetch)
 
+		case <-expire.C:
+			// The oldest fetch has timed out; the sweep at the top of the
+			// loop removes it and releases the peer's allowance.
+
 		case req := <-f.filter:
 			// Blocks arrived, extract any explicit fetches, return all else
 			var blocks []*nom.DetailedMomentum
@@ -394,6 +403,23 @@ func (f *Fetcher) reschedule(fetch *time.Timer) {
 		}
 	}
 	fetch.Reset(arriveTimeout - time.Since(earliest))
+}
+
+// rescheduleExpiry resets the expiry timer to the moment the oldest pending
+// fetch times out, so expired fetches are swept even while the loop is idle.
+func (f *Fetcher) rescheduleExpiry(expire *time.Timer) {
+	// Short circuit if no fetches are pending
+	if len(f.fetching) == 0 {
+		return
+	}
+	// Otherwise find the oldest fetch still waiting on a reply
+	earliest := time.Now()
+	for _, announce := range f.fetching {
+		if earliest.After(announce.time) {
+			earliest = announce.time
+		}
+	}
+	expire.Reset(fetchTimeout - time.Since(earliest))
 }
 
 // enqueue schedules a new future import operation, if the block to be imported
--- a/protocol/fetcher/fetcher.go
+++ b/protocol/fetcher/fetcher.go
@@ -37,7 +37,7 @@ const (
 	fetchTimeout  = 5 * time.Second        // Maximum alloted time to return an explicitly requested block
 	maxUncleDist  = 7                      // Maximum allowed backward distance from the chain head
 	maxQueueDist  = 32                     // Maximum allowed distance from the chain head to queue
-	hashLimit     = 256                    // Maximum number of unique blocks a peer may have announced
+	HashLimit     = 256                    // Maximum number of unique blocks a peer may have announced; also the largest batch one fetch request names
 	blockLimit    = 64                     // Maximum number of unique blocks a per may have delivered
 )
 
@@ -267,8 +267,8 @@ func (f *Fetcher) loop() {
 		case notification := <-f.notify:
 			// A block was announced, make sure the peer isn't DOSing us
 			count := f.announces[notification.origin] + 1
-			if count > hashLimit {
-				log.Info("Peer exceeded outstanding announces", "peer", notification.origin, "hash-limit", hashLimit)
+			if count > HashLimit {
+				log.Info("Peer exceeded outstanding announces", "peer", notification.origin, "hash-limit", HashLimit)
 				break
 			}
 			// All is well, schedule the announce if block's not yet downloading
@@ -300,10 +300,15 @@ func (f *Fetcher) loop() {
 					announce := announces[rand.Intn(len(announces))]
 					f.forgetHash(hash)
 
-					// If the block still didn't arrive, queue for fetching
+					// If the block still didn't arrive, queue for fetching.
+					// The fetching entry is outstanding work for the chosen
+					// peer, so count it again: forgetHash just released the
+					// announce and will release this entry when the fetch
+					// completes or times out.
 					if f.getBlock(hash) == nil {
 						request[announce.origin] = append(request[announce.origin], hash)
 						f.fetching[hash] = announce
+						f.announces[announce.origin]++
 					}
 				}
 			}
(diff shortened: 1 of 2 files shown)


edgepillar:
Thanks for separating the accounting correction into this PR. I reviewed `eac4a2e`; recharging the selected peer when an announcement becomes a fetch appears to repair the counter drift.

This is a source-level review: I did not execute the tests or independently verify CI. Two remaining points:

**1. P2 — Expiry can still race with admission at the event-selection boundary**

The sweep runs before `select`. If the timeout passes after that sweep and both `expire.C` and a notification are ready, `select` can choose the notification first. A peer at `HashLimit` can therefore have its new announcement rejected against already-expired fetching entries. The following iteration releases that allowance. [Sweep and admission](https://github.com/zenon-network/go-zenon/blob/eac4a2e01b200dd683ce64a8e1de450a74d6a618/protocol/fetcher/fetcher.go#L237-L295).

This is a transient lost announcement, not a crash, negative-counter recurrence, or quota escape. The timer provides idle cleanup but does not guarantee expiry-aware admission.

Could we retain the timer and, before rejecting for capacity, recheck whether expired fetches should release the allowance? A next-expiry check could avoid an unnecessary additional table walk when nothing has expired.

The regression should exercise admission against expired-but-still-counted entries without first waiting for `expiredHook`. An admission/expiry helper with explicit time, or a controlled clock, would avoid relying on randomized `select` ordering.

**2. P3 — The post-expiry test still assumes the new hashes remain in `announced`**

After sending eight new notifications, the test requires zero fetching entries and exactly eight announced entries. A scheduler pause long enough for `arriveTimeout` can legitimately move accepted hashes into fetching, causing a false failure. [Assertions](https://github.com/zenon-network/go-zenon/blob/eac4a2e01b200dd683ce64a8e1de450a74d6a618/protocol/fetcher/fetcher_test.go#L630-L640).

Please consider checking the exact eight new hashes across `announced ∪ fetching`, plus absence of the expired original hashes, or controlling the clock/transition explicitly. Union checking handles this phase change; eliminating dependence on longer pauses also requires time control.

Both follow-ups appear confined to `protocol/fetcher/fetcher.go` and `protocol/fetcher/fetcher_test.go`. No wire-format, quota increase, or broader downloader redesign is requested.

0x3639:
Both points verified and addressed in 1da58da (plus 8342516, a small change gating the test hook's slice collection on the hook being set).

**1. Expiry versus admission at the `select` boundary.** Confirmed. The state you describe, timed out but not yet swept when a notification wins the `select`, cannot be reached by the sweep-then-block structure alone. The notify path now calls `releaseExpired` for the announcing peer before counting: when that peer is at its allowance and the oldest pending fetch is past its timeout, the sweep runs first. The sweep already walks the table, so it now records the oldest survivor's announce time as it goes, and the admission check reads that field rather than walking anything. The sweep only runs from the notify path when at least one fetch has actually expired, and each fetch expires once, so a peer at its limit cannot turn refused announces into repeated walks. The timer stays for idle cleanup.

The regression test takes the helper route you suggested. It builds the exact between-sweep-and-select state by hand on a fetcher with no loop: `HashLimit` fetches from one peer, all past `fetchTimeout`, counter at the limit. `releaseExpired` frees them and the next announce would be admitted. A control test checks that live fetches are left alone, the peer stays refused, and a peer under its limit triggers no sweep even with stale fetches. Neither depends on `select` ordering or on waiting for the expiry hook. One honest gap: the one-line call site in the notify case is not itself pinned, since the limit check stays inline so the rename hunk shared with #84 still merges cleanly in either order (re-verified with `git merge-tree` and a race run of the merged tree).

**2. Post-expiry assertion.** Confirmed, same shape as the earlier lifecycle assertion. The test now checks that exactly the eight new hashes are held across `announced ∪ fetching` and that none of the expired originals remain. A one-second pause inserted after the new announces, long enough for them to move to fetching, passes.

Also since your review: the two timeout tests wait on an `expiredHook` the sweep calls after removing entries, in the same style as `fetchingHook`, rather than sleeping past `fetchTimeout`; and the sweep and the timer's oldest-survivor search are one walk rather than two.

edgepillar:
Thanks for the follow-up. I reviewed `1da58da`, including the intervening `8342516`, and consider both previously reported points addressed in the source.

The notify path now calls `releaseExpired` before calculating the admission count, covering the expired-but-still-counted state at the selection boundary. The lifecycle assertion now checks the exact expected hashes across `announced ∪ fetching`, excluding the expired originals. [Admission](https://github.com/zenon-network/go-zenon/blob/1da58da61b2dd500132ba80303e0bee363af9072/protocol/fetcher/fetcher.go#L272-L278), [assertions](https://github.com/zenon-network/go-zenon/blob/1da58da61b2dd500132ba80303e0bee363af9072/protocol/fetcher/fetcher_test.go#L641-L669).

I also agree with the acknowledged coverage qualification: the direct tests exercise the helper, not its placement in the notify path. Pinning that call site would be useful optional regression hardening, rather than evidence of a remaining production defect. The union assertion removes the phase-transition assumption without making the real-time test independent of arbitrarily long pauses.

I found no new blocking issue in this follow-up. This was a static review; I did not execute the tests or independently verify CI.

openmac0815:

Re-derived the fix from the diff against `dev`. Confirmed: increment of the chosen peer's counter when the announce moves to `fetching`, the `<= 0` floor in both `forgetHash` decrement paths, the expiry timer armed from `oldestFetch`, the hook-gated `expiredHook` slice, and `releaseExpired` on the notify path so a peer at its limit is judged on live entries only. The accounting invariant (counter covers announced + fetching, returns to zero at end of every lifecycle) is now actually enforced, and the tests observe sweeps via hooks rather than sleeps — deterministic.

One nit on presentation only: the scenario names in my own notes didn't all match the actual test function names (`TestAnnounces_ExactThroughFetchAndCompletion`, `TestAnnounces_FetchingCountsTowardHashLimit`, etc.) — the tests themselves cover exactly the right cases.

Release-critical for P2P sync health; no wire-protocol change; merge-compatible with #84 via the `HashLimit` export (verified either order).

PR #97, opened 2026-09-06
p2p/discover: expire node database entries without a node record

0x3639:
Follow-up to #86, addressing the retention point raised in review there. The behaviour is pre-existing on `dev` and independent of the admission change, so it is fixed separately.

### Problem

`Table.ping` records the ping time before the pong arrives. When the exchange fails, no node record is ever written, but the timing key stays. `expireNodes` only considered node-record keys, so these entries were never reclaimed. Every unanswered bond, whether from an unsolicited ping or from a lookup reaching a peer that has gone away, left a permanent key in the node database.

Growth is bounded by the bonding rate (16 slots, 500 ms timeout, so at most 32 identities per second) but unbounded in total. It also lengthens the hourly expiration scan and the linear seed scan used when the table is empty.

### Change

`expireNodes` now considers an identity whenever any discovery field is stored for it. The retention rule is unchanged: an identity is deleted when its last pong is older than the expiration window, which for an identity that was never answered is immediately. Nothing reads the ping time back, so no known-peer state is lost, and node records with a recent pong are kept exactly as before. Fields of one identity are adjacent in key order, so each identity is decided once per scan.

### Tests

`p2p/discover/database_test.go` (new): a stale ping-only identity is reclaimed, a node record with a recent pong is kept, and a node record with a stale pong is reclaimed as before. The first check fails on the previous code.

### Verification

- `GOWORK=off go vet ./p2p/discover/`
- `GOWORK=off go test -race -count=3 ./p2p/discover/`
- `GOWORK=off go test ./...`
- `GOWORK=off golangci-lint run ./p2p/discover/` (no new findings)

### DIFF
--- a/p2p/discover/database.go
+++ b/p2p/discover/database.go
@@ -24,6 +24,7 @@ import (
 	"encoding/binary"
 	"fmt"
 	"os"
+	"strings"
 	"sync"
 	"time"
 
@@ -248,6 +249,11 @@ func (db *nodeDB) expirer() {
 
 // expireNodes iterates over the database and deletes all nodes that have not
 // been seen (i.e. received a pong from) for some alloted time.
+//
+// An identity is considered whenever any discovery field is stored for it,
+// not only when it has a node record: a ping that was never answered leaves
+// timing metadata alone, and those entries expire on the same terms as a
+// node record whose last pong is old.
 func (db *nodeDB) expireNodes() error {
 	threshold := time.Now().Add(-nodeDBNodeExpiration)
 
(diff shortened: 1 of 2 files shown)


edgepillar:
Thanks for separating this retention correction from #86. Reviewed at `b33962`; this is a source-only review, without independent test or CI execution.

Considering metadata-only identities under the existing last-pong policy looks correct. The adjacent-identity check also avoids repeatedly deciding the same identity during one scan. The new stale ping-only, fresh bonded, and stale bonded controls directly cover the original concern.

One inherited cleanup issue is worth addressing alongside the newly reachable bulk-deletion path: `deleteNode` creates an iterator without releasing it or checking its error. `expireNodes` also ignores deletion errors and returns success without checking its outer iterator. Could these iterators be released and their errors propagated? [Deletion helper](https://github.com/zenon-network/go-zenon/blob/b3396228985a320ed0586cf96f26945eecac06e6/p2p/discover/database.go#L209-L217), [expiry loop](https://github.com/zenon-network/go-zenon/blob/b3396228985a320ed0586cf96f26945eecac06e6/p2p/discover/database.go#L257-L288).

These omissions predate this PR; I am not claiming a measured leak or a newly introduced exploit.

For coverage, fresh ping-only and fresh pong-only controls would pin the missing-pong policy explicitly. The retained-node assertion could also check the actual node record and pong value, rather than only whether some key remains.

One operational qualification remains: automatic expiry still starts only after successful bonding. This patch corrects what a sweep visits; it does not change that existing bootstrap gate.

openmac0815:

The leak is real: `expireNodes` previously only considered the node-record field (`field == nodeDBDiscoverRoot`), so ping-only identities from unanswered bonds were never reclaimed. The fix (prefix match on any discovery field + decide-each-identity-once via key adjacency) is minimal and preserves the retention rule — an identity that never pongs has a zero last-pong and expires immediately; nothing reads ping time back, so no live state is lost.

Note: the PR adds a single combined test (`TestExpireNodesReclaimsEntriesWithoutNodeRecord`) covering the ping-only/recent-pong/stale-pong scenarios — the coverage is right; early review notes listing three separate test names were inaccurate about names, not substance.

PR #98, opened 2026-09-06
node: stop the WebSocket RPC server on shutdown

0x3639:
Follow-up to a review finding on #87. Pre-existing and independent of that change.

## Problem

`httpServer.doStop` in `node/rpcstack.go` loads both handler variables from the HTTP handler slot:

```go
httpHandler := h.httpHandler.Load().(*rpcHandler)
wsHandler := h.httpHandler.Load().(*rpcHandler)
```

The WebSocket `rpc.Server` never receives its `Stop()` call. `Stop()` is what closes the codecs behind established WebSocket connections. Those connections are hijacked from `net/http` during the upgrade, so `http.Server.Shutdown` and closing the listener do not reach them. A client connected over WebSocket keeps an open, working connection after `stopRPC` returns. The WebSocket handler slot is also only cleared as a side effect when an HTTP handler happens to be installed on the same server.

In practice the process exits shortly after `stopRPC`, which is why this went unnoticed.

## Fix

Read the WebSocket handler from its own slot. One line.

## Test

`TestHTTPServerStopClosesWebSocketConnections` starts an `httpServer` with WebSocket enabled, opens a WebSocket client, proves it live with one JSON-RPC round trip, stops the server, then checks that the client's next read fails with a non-timeout error and that both handler slots are cleared. It covers a WebSocket-only server and one sharing its listener with HTTP, since the two shapes took different paths through the old code. Both subtests fail before the fix by timing out on the read.

```
GOWORK=off go test ./node/ -count=3   # ok
```

### DIFF
--- a/node/rpcstack.go
+++ b/node/rpcstack.go
@@ -240,7 +240,7 @@ func (h *httpServer) doStop() {
 
 	// Shut down the server.
 	httpHandler := h.httpHandler.Load().(*rpcHandler)
-	wsHandler := h.httpHandler.Load().(*rpcHandler)
+	wsHandler := h.wsHandler.Load().(*rpcHandler)
 	if httpHandler != nil {
 		h.httpHandler.Store((*rpcHandler)(nil))
 		httpHandler.server.Stop()
--- a/node/rpcstack_test.go
+++ b/node/rpcstack_test.go
@@ -1,0 +1,102 @@
+package node
+
+import (
+	"testing"
+	"time"
+
+	"github.com/gorilla/websocket"
+
+	rpc "github.com/zenon-network/go-zenon/rpc/server"
+)
+
+// stackProbe is the only API registered by these tests.
+type stackProbe struct{}
+
+func (stackProbe) Ping() string { return "pong" }
+
+var stackAPIs = []rpc.API{{Namespace: "probe", Service: stackProbe{}, Public: true}}
+
+// startStackServer configures an httpServer on an ephemeral loopback port
+// with WebSocket enabled and, when withHTTP is set, HTTP JSON-RPC too.
+func startStackServer(t *testing.T, withHTTP bool) *httpServer {
+	t.Helper()
+	h := newHTTPServer(rpc.DefaultHTTPTimeouts)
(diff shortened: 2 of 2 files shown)


edgepillar:
Reviewed at `1c7d766`. This is a source-only review; I have not executed the tests or independently verified CI.

The handler-slot correction looks right: `doStop` now stops the actual WebSocket RPC server. The WS-only and shared-listener tests meaningfully cover connections already registered before shutdown. [Fix](https://github.com/zenon-network/go-zenon/blob/1c7d7660b415ae630b7190a2472f9094567e5e5b/node/rpcstack.go#L241-L251), [regression](https://github.com/zenon-network/go-zenon/blob/1c7d7660b415ae630b7190a2472f9094567e5e5b/node/rpcstack_test.go#L68-L100).

One **separate, pre-existing** lifecycle gap remains worth tracking: `ServeCodec` checks `run` before adding the codec, while `Stop` changes `run` and scans the codec set separately. A codec can pass the check, pause until the shutdown scan finishes, and then register too late to be closed by that scan. [Admission](https://github.com/zenon-network/go-zenon/blob/1c7d7660b415ae630b7190a2472f9094567e5e5b/rpc/server/server.go#L74-L88), [shutdown](https://github.com/zenon-network/go-zenon/blob/1c7d7660b415ae630b7190a2472f9094567e5e5b/rpc/server/server.go#L121-L128).

Could that be tracked separately with a barrier-controlled registration-versus-`Stop` regression? Pause after the running-state check, complete shutdown, then resume registration and verify the codec cannot remain open.

This does not invalidate the one-line fix or require broadening this PR. It limits the shutdown guarantee: the current tests establish closure of already-registered connections, not every concurrent-upgrade ordering.

openmac0815:

Confirmed against the diff: `doStop` loaded both handler variables from `h.httpHandler`, so the WebSocket `rpc.Server` never got `Stop()` and hijacked WS connections survived shutdown. The one-line slot fix is correct, and the regression test proves actual connection closure (non-timeout read error) for both the WS-only server and the shared-listener shape — both fail on unpatched code by hanging.

Minimal risk, no behavior change for HTTP-only servers.

PR #101, opened 2026-09-19
Add community spork address renewal window

0x3639:
## Summary

`CommunitySporkAddressEndHeight` (13243712) was reached on mainnet on 2026-05-13, so the community spork address can no longer create or activate sporks.

This keeps the original window untouched and adds a **second, future renewal window**, so historical execution results are preserved and old/new binaries agree on every community spork call until the renewal window opens.

| Constant | Height | Target (observed ~11.35 s/momentum) | At nominal 10 s |
|---|---|---|---|
| `CommunitySporkAddressStartHeight` | 10109240 | unchanged | |
| `CommunitySporkAddressEndHeight` | 13243712 | unchanged | |
| `CommunitySporkAddressRenewalStartHeight` | **14455739** | 2026-10-19 12:00 UTC | ~2026-10-15 |
| `CommunitySporkAddressRenewalEndHeight` | **19791986** | 2028-09-19 12:00 UTC | ~2028-06-24 |

Projected from frontier 14228943 at 2026-09-19 16:57:40 UTC. The renewal start is ~30 days out to give operators time to upgrade.

Also:
- Version bumped to **v0.0.9** so the master release workflow publishes a new `v0.0.9-alphanet` tag instead of overwriting the `v0.0.8-alphanet` assets.
- New `TestSpork_CommunitySporkRenewalWindow` covers both boundaries of each window and the gap between them, evaluated at contract receive height (send frontier + 1), not send height.

## Mainnet history check

Scanned all 1104 account blocks of `z1qqvwzz2xq7q5gwk6uhcddgrpxlfcyzc8rsu82s` via public nodes. There is exactly one send to the spork contract in its history: the "Governance Contract" create, received at momentum 10231257 (inside the original window, succeeded). No community spork calls exist in the expired gap, so no historical contract receive changes under this PR.

## Rollout

All node operators must be running v0.0.9 before momentum **14455739**. Before that height both binaries produce identical results; after it, any community spork call will be rejected by v0.0.8 nodes and accepted by v0.0.9 nodes.

## Testing

- `go build ./...`
- `go test ./vm/...` passes (includes the full embedded tests package)


edgepillar:
Thanks for documenting the motivation and height calculation. I reviewed `3114e5d` statically. Before merging, could we address the following consensus and rollout concerns?

### 1. Preserve historical execution results

Changing only the end height retroactively authorizes community calls with an execution-context height in `[13243712, 19791731)`.

The height check runs in `ReceiveBlock`, not `ValidateSendBlock`. An otherwise valid community send can therefore be admitted under the old rules, with its contract receive recording a permission failure. Under the extended window, replaying that same call can succeed and produce different state and receive-block data.

The VM regenerates contract receives and compares their changes hash and block hash, so this can prevent historical synchronization if such transactions exist. This is not merely a difference in the returned error message. [Authorization paths](https://github.com/zenon-network/go-zenon/blob/667a69d9e9a418edf7580b08492ba5dcb9efd63a/vm/embedded/implementation/spork.go#L30-L165), [receive validation](https://github.com/zenon-network/go-zenon/blob/667a69d9e9a418edf7580b08492ba5dcb9efd63a/vm/vm.go#L90-L113), [failure finalization](https://github.com/zenon-network/go-zenon/blob/667a69d9e9a418edf7580b08492ba5dcb9efd63a/vm/vm.go#L195-L272).

I have not checked mainnet transaction history, so I am not claiming that an affected historical call exists. That audit would need to include failed Create/Activate receives, not just the current spork list.

A safer approach would preserve the original authorization window and add a separate, coordinated future renewal window. Moving only the original start height would instead invalidate earlier authorizations.

Even with no affected historical transactions, old and upgraded nodes will execute newly authorized calls differently. The six-momentum spork enforcement delay does not resolve this earlier authorization disagreement.

### 2. Add regression coverage for the actual rule change

The existing community tests replace the production heights with small test values, so passing them does not establish compatibility of this extension. [Existing tests](https://github.com/zenon-network/go-zenon/blob/667a69d9e9a418edf7580b08492ba5dcb9efd63a/vm/embedded/tests/spork_test.go#L51-L58).

Suggested deterministic coverage:

- Replay previously failed Create/Activate receives with unchanged result data, changes hash and block hash.
- Preserve authorization in the original window and rejection in the expired gap.
- Check both boundaries of any renewal window.
- Cover delayed receives: the relevant height is the source send’s confirmation height through `MomentumAcknowledged`, not wall-clock time or simply the latest frontier.
- Preserve rejection of unauthorized senders and the existing original spork-address behavior.

### 3. Make the release and authority-renewal plan explicit

The current master workflow publishes on master pushes, derives its tag from the unchanged `v0.0.8` version, and enables asset overwriting. This is existing infrastructure, but it could distribute different consensus rules under the same release identity. Please use a distinct, source-pinned release and an explicit operator upgrade/activation plan. [Release configuration](https://github.com/zenon-network/go-zenon/blob/667a69d9e9a418edf7580b08492ba5dcb9efd63a/.github/workflows/znn_builder.yml#L114-L127).

The renewed authority also retains the ability to activate an unimplemented spork, which intentionally terminates affected nodes. This is an existing privileged capability, not a newly discovered unauthenticated bypass, but it makes the renewal and rollout policy important. [Enforcement behavior](https://github.com/zenon-network/go-zenon/blob/667a69d9e9a418edf7580b08492ba5dcb9efd63a/chain/momentum_pool.go#L52-L68).

### Minor calculation correction

From the stated frontier, `19791731 - 14227961 = 5563770` momentums. At 10 seconds each, that is approximately 643.95 days from September 19, 2026—around June 24, 2028, rather than May. The September target remains a rate-dependent estimate, not a timestamp-enforced expiry.

This review is source-only; I have not independently run the build/tests or verified the reported mainnet samples.

0x3639:
Thanks for the thorough review. Agree on all points; the PR has been reworked (commit `1dc8fe3`, title and description updated).

**1. Historical execution results.** Went with your two-window suggestion. The original `[10109240, 13243712)` window is untouched and a separate renewal window `[14455739, 19791986)` is added, with `checkCommunitySporkAddressValidity` accepting either. The renewal start is ~30 days out so operators can upgrade before it opens; until then old and new binaries execute every community spork call identically.

I also did the audit you flagged as outstanding: scanned all 1104 account blocks of the community spork address on two public nodes. There is exactly one send to the spork contract ever, the "Governance Contract" create, received at momentum 10231257 inside the original window with a successful receive. Nothing in the expired gap, so no historical receive would have changed under either approach, but the two-window design removes the dependency on that fact.

**2. Regression coverage.** Added `TestSpork_CommunitySporkRenewalWindow`. It pins all four heights (original `[10, 15)`, renewal `[20, 25)`) and asserts, at contract receive height rather than send height: accepted inside the original window, rejected at the original end height, rejected in the gap, accepted at the renewal start height, accepted at the last renewal height, rejected at the renewal end height. The existing community tests now pin the renewal heights too so they no longer depend on the production values. Replaying historical receives isn't something the mock framework can do; the real check for that is a full mainnet resync from genesis with the new binary, which I'll run before this ships.

**3. Release / rollout.** Version bumped to v0.0.9 so the master workflow publishes `v0.0.9-alphanet` rather than overwriting the v0.0.8 assets. Rollout requirement is stated in the PR body: all nodes on v0.0.9 before momentum 14455739.

**Calculation.** You're right, at the nominal rate the end lands around 2028-06-24, not May. The constants' comments and the PR body now give both the observed-rate target and the nominal-rate date.

edgepillar:
Thanks for the revision. I checked `1dc8fe3`: preserving the original window and introducing a separate future renewal window addresses my original retroactive-authorization concern. The version bump also separates this release from the existing v0.0.8 assets.

A few focused follow-ups before shipping:

1. **Activation coverage and height terminology.** The new renewal test exercises `CreateSpork`, but not `ActivateSpork` across the renewal boundaries and expired gap. Although both methods share the authorization helper, an activation regression using existing inactive sporks would cover that integration path. [Current test](https://github.com/zenon-network/go-zenon/blob/1dc8fe318abf9a17fb05fc10da7263cd7eb01fae/vm/embedded/tests/spork_test.go#L125-L209).

   Please also clarify “frontier at receive time (send frontier + 1)” in the test comment and description. The protocol binds the receive’s `MomentumAcknowledged` to the source send’s **confirmation height**; “send frontier + 1” describes this fixture, not a general rule. A delayed-receive case would make that distinction explicit. [Context construction](https://github.com/zenon-network/go-zenon/blob/1dc8fe318abf9a17fb05fc10da7263cd7eb01fae/vm/supervisor.go#L343-L365).

2. **Release commit identity.** The version bump fixes the v0.0.8 asset collision, but both upload steps still omit `target_commit`. Could the release explicitly pin the tag to `${{ github.sha }}`, or document and verify an equivalent procedure ensuring the tag identifies the exact commit used to build the binaries? This can be handled as a coordinated release-workflow change rather than expanding the spork logic. [Release workflow](https://github.com/zenon-network/go-zenon/blob/1dc8fe318abf9a17fb05fc10da7263cd7eb01fae/.github/workflows/znn_builder.yml#L114-L190).

3. **Completed validation evidence.** Thanks for reporting the history audit and planning a genesis resync. Once the resync finishes, please record the exact build SHA, validated frontier height/hash and outcome alongside the test results, so the evidence is tied to the release candidate.

These are remaining test/documentation and release-verification items, not a repetition of the original historical-authorization objection. My review remains source-only; I have not independently run the tests, history scan or resync.

digitalSloth (commented):
Looks good, I think its almost there, just two things around tests and comments in the code that I noticed:

**1. Activate path is untested in the renewal window** (`vm/embedded/tests/spork_test.go:334-338`). `TestSpork_ActivateCommunitySpork` pins the renewal window to 100/200 and never reaches 100. Removing the renewal branch from `checkCommunitySporkAddressValidity` leaves the suite green. Please set a reachable range (e.g. 30/45) and add create+activate cases at `RenewalStart`, `RenewalEnd - 1`, `RenewalEnd`, and one gap height.
**2. Test comment names the wrong height** (`spork_test.go:121-124`). "Frontier at receive time (send frontier + 1)" is only true in the mock. The rule uses the momentum that confirmed the send (`MomentumAcknowledged`, `vm/supervisor.go:383-397`, enforced at `verifier/account_block.go:321-327`). A send confirmed at 14455738 is denied permanently even if its receive is generated later. Please fix the comment.

**Suggestion:** record the rollout constraint next to the heights in `vm/embedded/definition/spork.go:41-47`. Nodes on v0.0.8 and v0.0.9 diverge only if a community spork send is confirmed inside `[14455739, 19791986)`, and un-upgraded nodes then reject the momentum. So the gate is "do not broadcast a community spork tx until the network runs v0.0.9", not "all nodes on v0.0.9 before 14455739".

**Follow-up, not blocking:** the release workflow sets no target commit on `svenstaro/upload-release-action`, so `v0.0.9-alphanet` tags default-branch head rather than the built SHA

0x3639:
Addressed in `f0e8438`.

**Activate coverage (edgepillar 1, digitalSloth 1).** `TestSpork_CommunitySporkRenewalWindow` now exercises `ActivateSpork` at every boundary alongside `CreateSpork`: activation of an existing spork is denied when the send is confirmed at the original end height (15) and in the gap (19), accepted at the renewal start (20) and the last renewal height (24), and denied at the renewal end height (25). The accepted activations set enforcement heights 26 and 30 and are marked implemented so the mock node survives crossing them. One correction to digitalSloth's note: removing the renewal branch did not leave the suite green at `1dc8fe3`, the create-path assertions in the same test already failed. What was missing was the activate path, and the mutation now fails on both.

**Height terminology (edgepillar 1, digitalSloth 2).** The comment now states the rule as the send's confirmation height: the contract receive is built from the momentum store at its `MomentumAcknowledged`, which the supervisor sets to the confirming momentum and the verifier enforces. "Send frontier + 1" is called out as a property of the fixture only. The mock generates contract receives in the momentum after confirmation, so a delayed-receive case is not something it can express; the comment fix and the PR description carry that distinction instead.

**Rollout constraint (digitalSloth suggestion).** Recorded next to the renewal heights in `vm/embedded/definition/spork.go` and in the PR description: divergence happens only if a community spork send is confirmed inside the renewal window while un-upgraded nodes are validating, so the hard rule is "do not send a community spork transaction until the network runs v0.0.9". The "upgrade before 14455739" deadline is kept as the operational target that makes the window usable when it opens.

**Release commit identity (edgepillar 2, digitalSloth follow-up).** Folded into this PR rather than a separate one: both `upload-release-action` steps now pass `target_commit: ${{ github.sha }}`. The workflow publishes on the merge to master and an existing tag is never re-pointed, so a separate PR would have had to land first for `v0.0.9-alphanet` to benefit. The second step's `target_commit` is ignored once the tag exists, it is there so both steps read the same.

**Resync evidence (edgepillar 3).** Will be posted here once the genesis resync on the release candidate completes, with build SHA, validated frontier height and hash, and outcome.

Verification: `GOWORK=off go build ./...` and `GOWORK=off go test ./vm/... -count=1` pass on `f0e8438`.

0x3639:
@georgezgeorgez this is a consolidated summary of the PR and the review thread so it can be evaluated and tested without reading the back and forth.

The community spork address `z1qqvwzz2xq7q5gwk6uhcddgrpxlfcyzc8rsu82s` lost its authority to create or activate sporks when mainnet passed `CommunitySporkAddressEndHeight` (13243712) on 2026-05-13. This PR does not move that end height. It leaves the original window `[10109240, 13243712)` untouched and adds a second renewal window `[14455739, 19791986)`; `checkCommunitySporkAddressValidity` now accepts a call whose height falls in either window and denies it everywhere else, including the gap between them. The check runs in the contract receive at the height of the momentum that confirmed the send (the receive's `MomentumAcknowledged`), not the send's own height and not the frontier when the receive is generated. Because the original window is unchanged, every historical community spork receive replays with the same result, so v0.0.8 and v0.0.9 binaries agree at every height until the renewal window opens. A scan of all 1104 account blocks of the community address found exactly one send to the spork contract ever, the "Governance Contract" create received at momentum 10231257 inside the original window, so nothing in the expired gap would have changed under either design. The version is bumped to v0.0.9 so the master release workflow publishes `v0.0.9-alphanet` instead of overwriting the v0.0.8 assets, and both upload steps pass `target_commit: ${{ github.sha }}` so the tag identifies the built commit.

Rollout: v0.0.8 and v0.0.9 nodes diverge only if a community spork send is confirmed inside `[14455739, 19791986)`. A v0.0.8 node regenerates that receive with a permission failure, computes a different changes hash and rejects the momentum. The hard rule is therefore that no community spork transaction is sent until the network runs v0.0.9, and every node operator should be on v0.0.9 before momentum 14455739 (projected around 2026-10-19 at the observed ~11.35 s/momentum, around 2026-10-15 at the nominal 10 s) so the window is usable when it opens. This is recorded next to the heights in `vm/embedded/definition/spork.go`.

The [regression test](https://github.com/0x3639/go-zenon/blob/f0e8438830e2a9db3915832c30e6ce031752cfca/vm/embedded/tests/spork_test.go#L121) pins the windows to `[10, 15)` and `[20, 25)` and exercises both `CreateSpork` and `ActivateSpork` at each boundary, evaluated at the send's confirmation height: accepted at 11, denied at 15 (original end, exclusive), denied at 19 (gap), accepted at 20 (renewal start, inclusive), accepted at 24, denied at 25 (renewal end, exclusive). Verified with Go 1.25.4: it fails against the old authorization logic at `667a69d`, where the create confirmed at renewal start height 20 is denied so the test reports `spork spork-renewal-start not found` and the expected `created`/`activated` log lines are missing, and it passes at PR head `f0e8438`. `GOWORK=off go build ./...` and `GOWORK=off go test ./vm/... -count=1` also pass at `f0e8438`. Run on the PR branch:

```sh
GOWORK=off go test ./vm/embedded/tests -run '^TestSpork_CommunitySporkRenewalWindow$' -count=1
```

<details>
<summary>Reproduce the failure before the fix</summary>

From a checkout containing the PR commit, run:

```sh
pr=f0e8438830e2a9db3915832c30e6ce031752cfca
before=$(mktemp -d)
git archive 667a69d9e9a418edf7580b08492ba5dcb9efd63a | tar -x -C "$before"
git show "$pr":vm/embedded/tests/spork_test.go > "$before/vm/embedded/tests/spork_test.go"
git show "$pr":vm/embedded/definition/spork.go > "$before/vm/embedded/definition/spork.go"
(cd "$before" && GOWORK=off go test ./vm/embedded/tests -run '^TestSpork_CommunitySporkRenewalWindow$' -count=1)
```

The temporary copy includes the unchanged regression test and the definition file, which only declares the two renewal height variables and their comments so the test compiles. The authorization logic in `vm/embedded/implementation/spork.go` stays at the base. The run should fail as described above. Your checkout is untouched.

</details>

Review history: edgepillar's original objection, that extending the end height would retroactively authorize calls in the expired gap and could break historical sync, was resolved by the two-window design in `1dc8fe3`. The activate-path coverage, `MomentumAcknowledged` wording, rollout note and `target_commit` follow-ups from edgepillar and digitalSloth landed in `f0e8438`. What remains: the mock framework generates contract receives in the momentum after confirmation, so it cannot express a delayed receive; the test comment and the PR description carry that rule instead. The genesis resync of the release candidate that edgepillar asked for (build SHA, validated frontier height and hash, outcome) has not been posted yet and will be added here when it completes. Both reviews were source-only; neither reviewer has run the tests, history scan or resync.

0x3639:
**Genesis resync evidence (edgepillar 3).** Built from [`2d6fdf3`](https://github.com/0x3639/go-zenon/commit/2d6fdf37aaee7b701a3d120f6471ac4c2edd1458) on branch `resync/pr81-pr101` of the 0x3639 fork. That commit is master `667a69d` plus the merge of this PR at `f0e8438` and PR #81 at `2be586f`, nothing else; the two PRs touch disjoint files and the binary reports v0.0.9. A full mainnet resync from genesis with that build completed without error.

edgepillar:
Thanks for reporting the completed genesis resync with `2d6fdf3`. Could you add the final validated frontier momentum height and hash? That would complete the build SHA / frontier / outcome record requested above.

Noted that this run covers the combined #81 + #101 build. My review remains source-only; I have not independently reproduced the resync.

0x3639:
> Thanks for reporting the completed genesis resync with `2d6fdf3`. Could you add the final validated frontier momentum height and hash? That would complete the build SHA / frontier / outcome record requested above.
> 
> Noted that this run covers the combined #81 + #101 build. My review remains source-only; I have not independently reproduced the resync.

Momentum Hash:
`91adf783541ae4b85eeb98172b9e8f65d8a7019a40537c7f21121fd7ece1b8c0`

Momentum Height:
`14,246,715`

openmac0815:

Verified against the diff: original window `[10109240, 13243712)` untouched, renewal window `[14455739, 19791986)` added as a second independent range, authorization evaluated at the confirmation momentum's height (the receive's frontier), gap between windows preserves current denial behavior — so old and new binaries agree on every historical call and divergence is only possible on a community-spork send confirmed inside the renewal window. Test covers both boundaries of both windows plus the gap, for create and activate. Version bump to v0.0.9 and `target_commit` pinning in the release workflow are correct.

**Operational flag:** the renewal opens at momentum 14,455,739 (~2026-10-19). Every operator must be on a release containing this window before then; the first community-spork send confirmed inside the window forks any un-upgraded node. Recommend cutting v0.0.9 well ahead of that date and communicating the deadline to pillars.

(For correctness the mainnet-history scan is actually redundant — the renewal window lies entirely in the future, so no historical receive can change — but it's good due diligence.)

